PR #60, opened 2018-02-28
Remove "dfn-" from W3C links

sjetha-pason:
The "dfn-" prefix for all anchor points linked to in this page does not take the user to the actual anchor on the page, since the "dfn-" has been dropped. This change removes all of the "dfn-" prefixes from all links to anchors so that the actual section of the documentation is shown when the user clicks the link.

I did a find and replace to remove all of the links and spot-checked a few of them.


erikadoyle:
Thanks for the corrections @sjetha-pason!

PR #62, opened 2018-03-07
Extension for Enterprise Updates

davidwesst:
Hello!

I've made some suggested updates to the extensions for enterprise documentation as there were a few things I believe should be clarified that are unique to developing extensions for enterprise.

Feedback back is welcome.

Thanks!


erikadoyle:
This looks great! Thanks @davidwesst

PR #51, opened 2017-12-04
Adding new documentation for the Microsoft Edge DevTools Protocol

pelavall:
Adding new documentation for the Microsoft Edge DevTools Protocol. This should **not** be merged without a review from @brendyna. I believe there are still files/articles which need to be added. Opening PR for comments on existing files and to get the process started.

PR #65, opened 2018-03-20
Adding new documentation for the Microsoft Edge DevTools Protocol - backup

pelavall:
Adding new documentation for the Microsoft Edge DevTools Protocol. This should not be merged without a review from @brendyna. I believe there are still files/articles which need to be added. Opening PR for comments on existing files and to get the process started.


pelavall:
Covered by #51.

PR #88, opened 2018-05-11
More on MSWebView*NavigationStarting events

david-risney:
Adding some extra / more explicit info about the navigationStarting events and fixing the navigationStarting event text.


erikadoyle:
Thank you @david-risney! I'll merge it live today.

PR #107, opened 2018-05-26
webview.NavigationCompletedEvent.uri property

david-risney:
Add missing uri property from base class NavigationEvent


erikadoyle:
Thanks so much @david-risney !

PR #113, opened 2018-06-09
Update WebDriver support

InstyleVII:
Updated the WebDriver docs to reflect the current support status and added missing capabilities and new command line parameters.


libbymc:
Thanks, @InstyleVII!

PR #116, opened 2018-06-11
Update webview permissionrequested eventargs

david-risney:


david-risney:
Not sure what I've done to make it think I've modified the entire file. Looking at the rich diff shows what I actually changed. Let me know if this is fine or if you have hints about what I might have done wrong.

microsoft-edge/webview/PermissionRequest.md
```diff
@@ -1,30 +1,140 @@
...
+ms.author: libbymc
+ms.date: 02/12/2018
+ms.topic: reference
+ms.prod: microsoft-edge
+keywords: webview, windows 10 apps, uwp, edge
+---
+
+# PermissionRequest object
+
+Provides information about a permission request. This object is available from the permissionRequest property of the event args from the [MSWebViewPermissioniRequested](../webview.md#MSWebViewPermissionRequested) webview event.
```
david-risney: This is the illegal link the PR tooling mentioned. Need to figure out what the correct way to do this is.

microsoft-edge/webview/PermissionRequestedEvent.md
```diff
@@ -0,0 +1,60 @@
+---
```
david-risney: New type doc added without corresponding entry in the toc

microsoft-edge/webview/PermissionRequest.md
libbymc: Just make the link lower case (../webview.md##mswebviewpermissionrequested) and that should work!

david-risney:
It was line endings like you suggested. I ended up needing to fix that so that it was mergeable for toc.md anyway. Thanks

PR #125, opened 2018-06-26
Added Known Issues sections to Peformance, Memory, and Network docs

zoherghadyali:
We needed links to add to four new error overlays we're improving in RS5:
- ProfilingStartError
- ProfilingStopError
- NetworkCollectionStartError
- NetworkCollectionStopError
I duplicated content for these because even though the work-around is the same, we can now provide specific links depending on which error the user encounters in the error overlay. Check this out as an example: https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide/memory#an-error-occurred-while-starting-the-profiling-session


erikadoyle:
@zoherghadyali, is this ready to push live now, or later for an upcoming flight?

zoherghadyali:
This could be pushed live now. Our tools won't point users to these sections until RS5 ships but at least it's good info for people to have if they encounter problems in the Network, Performance, or Memory tools.

PR #110, opened 2018-05-31
Updating EDP docs with RS5 changes

brendyna:
@erikadoyle, @pelavall Here's an update from running Pete's scripts and adding to the TOC. Mainly looking to get this in the staging environment so I can point folks to our RS5 docs (and get them prepped for RS5 release when it comes).

Other pages like Clients can come later as we nail down EDP. Trying to just get the APIs up for now.

PR #133, opened 2018-07-19
Updating EDP docs with RS5 changes

pelavall:
@erikadoyle, @brendyna Here's an update from domain generation scripts and adding to the TOC. Mainly looking to get this in the staging environment so I can point folks to our RS5 docs (and get them prepped for RS5 release when it comes).
Other pages like Clients can come later as we nail down EDP. Trying to just get the APIs up for now.

Reopening of #110 with branch in MicrosoftDocs/edge-developer


erikadoyle:
@pelavall I think I know the problem...looking into it

erikadoyle:
Most of the build warning should correct themselves when I merge to master. Will address the remaining ones from there (+ some final changes from Brendyn).

PR #136, opened 2018-07-22
Used === -1 instead of <= -1 in creating-an-extension.md

JoshuaKGoldberg:
`indexOf` will never return anything less than `-1`.


erikadoyle:
Thanks for catching that, @JoshuaKGoldberg !

PR #153, opened 2018-08-15
webView.addWebAllowedObject update

david-risney:
Additional info on addWebAllowedObject


erikadoyle:
Thanks @david-risney! Will push this live today.

PR #161, opened 2018-09-05
Updated shortcuts in the Debugger for a11y

zoherghadyali:
We needed to call out Ctrl+M in the Debugger as a way to escape focus trapping in the editor for accessibility. See: Microsoft/monaco-editor#296 for reference. I also just made the Console shortcuts look a little more consistent since I thought "" was confusing when we had two shortcuts that do the same thing.


erikadoyle:
Thanks @zoherghadyali

PR #162, opened 2018-09-06
Fix small typo in Microsoft Store PWA details page

cglong:


mattwojo:
Thanks again for catching this @cglong !

PR #168, opened 2018-09-17
Fix a broken bookmark in microsoft-store.md

yufeih:
bookmarks are case sensitive. docfx generate heading ids with lowercase characters.


erikadoyle:
Thank you @yufeih !

PR #187, opened 2018-10-22
fix: Replace RawGit with W3C TR accname page

nschonni:
RawGit is shutting down, so swapped the live version with the published w3c TR page


erikadoyle:
Thanks so much @nschonni !

PR #192, opened 2018-10-29
typo: publically -> publicaly

nschonni:


nschonni:
Opps, didn't drop the "a" should be "publicly"

PR #201, opened 2018-11-08
Branding update

JnHs:
Updated to use the new "Partner Center" branding. (TOC should ideally be updated to match topic title)


erikadoyle:
Thanks @JnHs !

PR #203, opened 2018-11-12
Fix formatting on Windows Feature article

ZhangYiJiang:
The page has a misindented blockquote which is causing the subsequent bullets to be incorrectly numbered and showing up as a code block.


mattwojo:
Thank you @ZhangYiJiang! I just merged your fix in to the live docs. We appreciate your help!

PR #241, opened 2019-03-24
fix: MD039/no-space-in-links

nschonni:
Spaces inside link text


jm-trd-ms:
#sign-off and and merge.

PR #245, opened 2019-03-24
fix: MD023/heading-start-left/header-start-left

nschonni:
Headings must start at the beginning of the line


jm-trd-ms:
#sign-off and merge.

PR #290, opened 2019-07-22
Remove feature flag of using v2 cache

v-pegao:
Feature flag `use_v2_github_user_cache` is only used during migrating to docfx v3, which can be removed now, it won't bring any impact to your content and publish result.


jm-trd-ms:
#sign-off and merge.

PR #319, opened 2019-09-11
WebView2 SDK 0.8.270 Add missing code samples

david-risney:
Fix various issues 
 * GettingStarted.md: Update the code in the GettingStarted document so it will actually compile after the changes to the WebView2 loader export.
 * IWebView2Environment2.md: Add some documentation to the IWebView2Environment2 interface which previously had none.
 * IWebView2HttpHeadersCollectionIterator.md: Add some documentation to the IWebView2HttpHeadersCollectionIterator interface which previously had none.
 * IWebView2WebView.md: Add back the missing sample code for the DocumentStateChanged event and an additional note about sandboxing for the AddScriptToExecuteOnDocumentCreated
 * IWebView2WebView3.md: Add back the missing sample code for the NewWindowRequested event.
 * webview2.idl.md: Replace smart quote with ASCII quote to avoid encoding issue. Use {} to represent variables to be consistent with existing documentation and avoid encoding issues with <>. Put command line switches in markdown code snippets so the double dash doesn't become ndash.


jm-trd-ms:
I don't see any immediate blockers for this to merge (all blockers are planned to be added in the IDL at a later date). 
#signoff

PR #331, opened 2019-10-01
Chromium docs 3709955

jm-trd-ms:


jm-trd-ms:
Merging PR from chromium.WebFundamentals plus updated doc file.

PR #372, opened 2019-11-12
Move ms.technology inside fileMetadata

yufeih:
The syntax of setting folder level metadata only takes effect when putting under `fileMetadata` section.


jm-trd-ms:
#sign off and merge.

PR #401, opened 2019-12-21
Updated screenshots and text about edge browser

flushpot1125:
Current edge browser seems to not support service worker well.

sw

It may makes students who learn PWA based on Visual Studio.  

The update is under the following environment.

[Microsoft Edge Beta Channel](https://www.microsoftedgeinsider.com/en-us/download/) 79.0.309.54


jm-trd-ms:
@flushpot1125 
Thank you for taking the time to submit these changes.
Your suggestion has been evaluated and may be implemented in future.

PR #414, opened 2020-01-09
Updated main Edge landing page

jm-trd-ms:
Replaced index.md with index.yml.


microsoft-edge/index.yml
```diff
@@ -0,0 +1,445 @@
+### YamlMime:Hub
+
+title: Microsoft Edge developer documentation
```
zoherghadyali: Do we need to say "developer" here? Our Enterprise content is more geared to IT pros and I'm sure we'll publish docs for other Edge customers as well. Can we change this to "Microsoft Edge documentation"

microsoft-edge/index.yml
```diff
@@ -0,0 +1,445 @@
...
+summary: Microsoft Edge is the browser built for Windows 10, providing fast, safe, and productive web experiences.  Here are all the developer features and tools available to you in Microsoft Edge and the Chromium engine.
```
zoherghadyali: This feels out of date. We build on macOS and ARM now too. We should probably get one of the writers/branding folks to come up with the correct verbiage to put here.

microsoft-edge/index.yml
```diff
@@ -0,0 +1,445 @@
...
+brand: m365
+metadata:
+  title: Microsoft Edge developer documentation
+  description: Microsoft Edge information for developers including PWAs, DevTools, extensions, webdriver, and more.
```
zoherghadyali: I don't think this is entirely accurate. We also include information for IT admins and regular browser users as well.

zoherghadyali: Also, "WebDriver" should be capitalized.

microsoft-edge/index.yml
```diff
@@ -0,0 +1,445 @@
...
+        - url: ./hosting/webview2/reference-webview2.md
+          itemType: reference
+          text: Reference
+    # Card
+    - title: Getting Started with DevTools
+      links:
+        - url: ./devtools-guide-chromium.md
+          itemType: reference
+          text: DevTools Overview
+        - url: ./devtools-guide-chromium/whats-new.md
```
zoherghadyali: This isn't correct any more. I've actually removed this doc from the toc.md file. The currently correct path would be: `./devtools-guide-chromium/whats-new/2019/12/devtools.md` and we'll just have to update this for each What's New release we do.

microsoft-edge/index.yml
```diff
@@ -0,0 +1,445 @@
...
+          url: ./webdriver.md
+        - text: Puppeteer
+          itemType: reference
+          url: (to be written)
+        - text: Interop (Site compatibility)
+          itemType: reference
+          url: ./web-platform/site-impacting-changes.md
+        - text: webhint
+          itemType: reference
+          url: (nothing written, but it would be good to link out to the browser extension)
```
zoherghadyali: You can probably link to the webhint user guide here: https://webhint.io/docs/user-guide/

microsoft-edge/index.yml
ststimac: @zoherghadyali coming here to +1 this. This should just be Microsoft Edge documentation.

microsoft-edge/index.yml
ststimac: I'm suggesting for now we change it to "Explore documentation resources for Microsoft Edge" and then we can work with marketing after 1/15.

microsoft-edge/index.yml
jm-trd-ms: Updated.
"Microsoft Edge information for developers, IT admins, and browser end users; including DevTools, Extensions, PWAs, WebDriver, and more."

microsoft-edge/index.yml
jm-trd-ms: Updated.

microsoft-edge/index.yml
jm-trd-ms: Updated.
"Explore documentation resources for Microsoft Edge.  Here are all the developer features and tools available to you in Microsoft Edge and the Chromium engine."

microsoft-edge/index.yml
```diff
@@ -0,0 +1,438 @@
...
+          url: ./accessibility.md
+        - text: DevTools Protocol (Chromium)
+          itemType: reference
+          url: ./devtools-protocol-chromium.md
+        - text: WebDriver
+          itemType: reference
+          url: ./webdriver.md  
+        - text: Puppeteer (need link)  
+          itemType: reference  
+          url: (to be written)
```
zoherghadyali: Please add this: https://docs.microsoft.com/en-us/microsoft-edge/puppeteer as the URL for Puppeteer.

PR #498, opened 2020-02-26
Added information regarding pipes in titles

BenjaminWalsh:
This information will help devs identify why reader view doesn't work when using pipes in page titles in edge.


jm-trd-ms:
Updated content per customer feedback.

jm-trd-ms:
#sign-off and merge.

PR #500, opened 2020-02-26
Minor change - wording

kizerkizer:
adjusted wording; initial version did not read well to me. Reject if you (docs folks) think otherwise.


jm-trd-ms:
@kizerkizer  
Thank you for taking the time to submit these changes.  
Your suggestion has been evaluated and may be implemented in future.

PR #509, opened 2020-03-04
Numbering and registry key location fixed for Native Messaging documentation

Znert:
Both headlines used "1." as numbering and the registry key location was wrong for the `HKEY_CURRENT_USER` case.


gebansal:
Thank you for your update. We have incorporated the changes in our latest update to the native messaging documentation. Please let us know if you see any concerns with it, would love your feedback. We are closing this PR for now.

PR #511, opened 2020-03-06
chore: Add Markdownlint config

nschonni:
Added the Markdownlint config by running `markdownlint "**/*.md"` and disabling all of the currently non-passing rules.
For the second 2 commits, I ran `markdownlint --fix "**/*.md"` after enabling those particular rules since they had a small amount of impact files.

If this lands a GitHub Action could be added like in the dotnet/docs https://github.com/dotnet/docs/blob/master/.github/workflows/markdownlint.yml


jm-trd-ms:
#sign-off and merge.

PR #512, opened 2020-03-09
Removed a note saying "Do not use the zip of a folder"

ashishnirvana1:
Removed the note as the developer can now zip a folder as well. Creating pull request.


jm-trd-ms:
#sign-off and merge.

PR #524, opened 2020-03-10
fix: Cleanup single value metadata

nschonni:
Ran Docs Authoring Pack "Cleanup: Single Value metadata"


nschonni:
Replaced by #538

PR #537, opened 2020-03-19
fix: HTTP -> HTTPS link

nschonni:
Replaces #525 because it had `>` in the branch name, that fails on windows checkout


jm-trd-ms:
Content updated per customer feedback.

jm-trd-ms:
#sign-off and merge.

PR #568, opened 2020-04-03
Updated index file

jm-trd-ms:
Added link to index.yml.


jm-trd-ms:
received #sign-off from requester offline.

PR #572, opened 2020-04-06
Chromium docs 4502624

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #573, opened 2020-04-06
Updated multiple extensions files

jm-trd-ms:
Updated content in multiple files to support media directory clean-up.


jm-trd-ms:
#sign-off received offline and merged.

PR #575, opened 2020-04-07
Updated webview content

jm-trd-ms:
Updated content to align with media directory clean-up.


jm-trd-ms:
Effort delayed until after //Build.

PR #577, opened 2020-04-08
Static versions of the GIFs in the old documents

codepo8:
Optimised with imageOptim.


jm-trd-ms:
Closing as duplicates of #pr560.

PR #586, opened 2020-04-10
Update accessibility.md

boggydigital:
Minor updates to exclude content only relevant for EdgeHTML and make it more appropriate for Chromium. Since we added UI Automation support to Chromium - content is still generally applicable, so just removing outdated references might be a good mitigation for now.


jm-trd-ms:
#sign-off and and merged.

PR #592, opened 2020-04-14
WebView2 rename BYO terms

JoshuaWeber:
DO NOT MERGE - Still under review with others, but want to see it on staging.

Update to the main WebView2 doc page. Primary work has been to update to the new runtime model by used and to rebrand the BYO scenario to use a more accurate fixed version.


JoshuaWeber:
> A single install of the [Evergreen] run-time will support any number of [Evergreen] WebView2 applications
> 
> Calling out evergreen here since BYO also includes WV runtime.

Is this actually a closed decision? I thought we were going to not use the WV runtime for BYO. And instead just allow the dev to package the binaries directly in their app.

microsoft-edge/hosting/WebView2.md
```diff
@@ -33,14 +33,29 @@ The [WebView2 Samples](https://github.com/MicrosoftEdge/WebView2Samples) reposit
 Help us build a richer WebView2 experience by sharing your feedback! Visit our [feedback repo](https://aka.ms/webviewfeedback) to submit feature requests or bug reports. It’s also a good place to search for known issues.
 During developer preview, we will also be collecting telemetry data to help us build a better WebView. Users can turn off WebView data collection by navigating to edge://settings/privacy in the browser and turning off browser data collection.
 
-## Evergreen vs. Bring-your-own
+## App Distribution
```
likuba, line 36: Wondering if this should just be 'Distribution' instead of 'App Distribution'?  Most of it is about how the WebView bits get distributed in the context of the app vs the app itself.

microsoft-edge/hosting/WebView2.md
```diff
@@ -33,14 +33,29 @@ The [WebView2 Samples](https://github.com/MicrosoftEdge/WebView2Samples) reposit
...
-In the future, there will be a second bring-your-own (BYO) option that allows developers to bundle a redistributable version of the browser with their apps. BYO brings a locked platform, but requires a larger disk footprint for the packaged browser and developers will have to take on the responsibility of servicing and updating the control themselves.
+The WebView2 control utilizes the Microsoft Edge (Chromium) browser. When distributing your app it is important to ensure that the Edge browser is installed on all user machines where your application will run. You should also be aware of which version and channel is installed, e.g. Stable, Beta, Dev, or Canary. The WebView2 controller will utilize the most stable version of the browser when installed on a machine.
```
likuba, line 38: My understanding is that stable channel will break in the upcoming SDK release next week.  So we may want to track a minor update to this paragraph to happen when that goes out?

microsoft-edge/hosting/WebView2.md
```diff
@@ -33,14 +33,29 @@ The [WebView2 Samples](https://github.com/MicrosoftEdge/WebView2Samples) reposit
...
+
+### Microsoft Edge WebView2 Runtime
+
+The Microsoft Edge WebView2 Runtime is a packaging of the browser binaries optimized for use by WebView2 applications. It will function stand alone or side-by-side with a client Edge Browser install. A single install of the run-time will support any number of WebView2 applications. Install of the runtime will not appear as a browser install to end-user, i.e. no desktop shortcuts, start menu entry, or protocol registration.
+
+An application utilizing WebView2 must ensure the installation of the Microsoft Edge WebView2 Runtime has occurred. At application install time you should check the current install status, which can be determined by using [GetCoreWebView2BrowserVersionInfo](https://docs.microsoft.com/microsoft-edge/hosting/webview2/reference/webview2.idl#getcorewebview2browserversioninfo). If the WebView2 Runtime is not available, you should chain the Microsoft Edge WebView2 Runtime Installer to your install flow.
```
likuba, line 54: i think some apps will choose to use existing fallbacks rather than chain the installer.  i am okay leaving this as-is with should, but wanted to call it out in case there are minor wording tweaks we want to consider

microsoft-edge/hosting/WebView2.md
pagoe-msft: Had the same thought. Can we add a sentence saying "temporarily we will not support the Stable channel"

liminzhu, line 38: I think there is still value in having a separate browser channel section, as it's not really about distribution, different channels are in general needed for different purposes. There is also important detail like the regkey that is lost here.

liminzhu: > When distributing your app it is important to ensure that the Edge browser is installed on all user machines where your application will run.

I think your intent is to capture the current state of the world here? Still, this is a pretty dangerous message and the very thing we want to get out of in the upcoming changes :). This is not technically true even today, as BYO binaries shouldn't considered as the browser.

microsoft-edge/hosting/WebView2.md
```diff
@@ -33,14 +33,29 @@ The [WebView2 Samples](https://github.com/MicrosoftEdge/WebView2Samples) reposit
...
+
+We also recognize that some app developers operate in constrained environments, and so intend to support two distribution options, evergreen and fixed version.
+
+Evergreen is the recommended distribution model for most developers. In this mode updates to the WebView2 Runtime will be fully managed by Microsoft keeping you up to date automatically without additional effort from your application. With this self-updating model you can be assured that your app is taking advantage of the latest features and security updates for hosted web content.
+
+For constrained environments we will also support a fixed version distribution model. In this model your application will select and package a specific WebView2 version. Updates to the WebView version will be the responsibility of the application and will be independent of any managed Microsoft update mechanisms. You should choose this model if it is crucial to be able to have absolute control over the browser version and update times your application is taking advantage of.
+
+### Microsoft Edge WebView2 Runtime
```
liminzhu, line 50: Microsoft Edge WebView2 Runtime is also not a reality today, should it be in part of the "future planned changes"?

microsoft-edge/hosting/WebView2.md
```diff
@@ -33,14 +33,29 @@ The [WebView2 Samples](https://github.com/MicrosoftEdge/WebView2Samples) reposit
...
+
+The Microsoft Edge WebView2 Runtime is a packaging of the browser binaries optimized for use by WebView2 applications. It will function stand alone or side-by-side with a client Edge Browser install. A single install of the run-time will support any number of WebView2 applications. Install of the runtime will not appear as a browser install to end-user, i.e. no desktop shortcuts, start menu entry, or protocol registration.
```
liminzhu, line 52: I want to be careful here that WV RT is not an evergreen thing, BYO uses the same WV RT binaries too. You might want to be clear that you're only talking about evergreen here.

liminzhu:
> A single install of the [Evergreen] run-time will support any number of [Evergreen] WebView2 applications

> Calling out evergreen here since BYO also includes WV runtime.

> Is this actually a closed decision? I thought we were going to not use the WV runtime for BYO. And instead just allow the dev to package the binaries directly in their app.

Well, BYO will definitely use the WV runtime (RT refers to the actual binaries that the installer installs, not the installer itself). What is being decided is whether BYO has to be installed on end user side through an installer, or can be packaged by developers as loose files.

PR #602, opened 2020-04-16
Chromium docs 4446128

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #654, opened 2020-05-05
Migrate Selenium tools for MS Edge documentation the MicrosoftDocs and include Python samples

bwalderman:
This change updates the WebDriver (Chromium) page in the Edge Developer docs to be the single destination for info on WebDriver integration for both Selenium 3 and 4. This includes info on the Selenium Tools for Microsoft Edge. A separate change will remove the duplicate content from https://github.com/microsoft/edge-selenium-tools and update the project to link here instead.

Now that the python package is close to being published, I've added Python samples. Comments have been added to emphasize that the samples are usable in bot Selenium 3 and 4.

### DIFF
--- a/microsoft-edge/webdriver-chromium.md
+++ b/microsoft-edge/webdriver-chromium.md
@@ -41,86 +41,85 @@ Now, [download the matching version of Microsoft Edge Driver][MicrosoftDeveloper
 > [!NOTE]
 > Microsoft Edge \(EdgeHTML\) does not work with [Microsoft Edge Driver][MicrosoftDeveloperEdgeToolsWebdriverDownloads].  To automate Microsoft Edge \(EdgeHTML\), you must download [Microsoft WebDriver for Microsoft Edge \(EdgeHTML\)][Webdriver].  
 
-## Download a WebDriver language binding  
+## Choose a WebDriver language binding  
 
 The last component you must download is a language-specific client driver.  The language binding translates the code you write in Python, Java, C#, Ruby, and JavaScript into commands that the Microsoft Edge Driver you downloaded in the [previous section](#download-microsoft-edge-driver) is able to run in Microsoft Edge \(Chromium\).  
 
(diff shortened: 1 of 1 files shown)


zoherghadyali:
Overall looks good but I have some more changes I'd like to make. For the Python and C# samples, I think we should actually use tabs like in this doc: https://docs.microsoft.com/en-us/office/dev/add-ins/quickstarts/outlook-quickstart?tabs=yeomangenerator. The way to create tabs is shown here: https://review.docs.microsoft.com/en-us/help/contribute/validation-ref/tabbed-conceptual?branch=master. 

I also think we should separate out the Selenium Tools for Microsoft Edge section into its own doc underneath WebDriver (Chromium). Then, from our Selenium 3 section, we can just point to it instead of embedding all of that content for Selenium Tools there.

I will try to get these changes in by EOD today but if you have the bandwidth, feel free to make them yourself and I'm happy to just review.

bwalderman:
@zoherghadyali thanks, I wanted to do tabs as well but wasn't aware how to actually create those. I'll take a look.

 > I also think we should separate out the Selenium Tools for Microsoft Edge section into its own doc underneath WebDriver (Chromium). Then, from our Selenium 3 section, we can just point to it instead of embedding all of that content for Selenium Tools there.

We're essentially doing this already. The Selenium 3 contains a couple of short paragraphs explaining why the Tools are needed, and the "Install Selenium Tools for Microsoft Edge" link points to the github repo where detailed install instructions can be found.

jm-trd-ms:
Based upon my reading about Tabs, it has been deprecated.
I will investigate further and respond here.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

bwalderman:
@jm-trd-ms, using Chocolatey or Docker to install WebDriver is independent of what language you choose to write your tests in so I think these should be left out of the tab group.

I was also imagining a separate tab group for each individual code sample. The current approach requires us to maintain two copies of the same content (one for each tab).

@jm-trd-ms, @zoherghadyali what are your thoughts?

jm-trd-ms:
#sign-off and merged.

PR #675, opened 2020-05-14
InPrivate, Extensions, Collections Update

lilliankravitz:
IME for InPrivate
Extensions Update
Collections pinterest update


jm-trd-ms:
#sign-off and merged.

Issue #665, opened 2020-05-13
typo


ciufudean:
"and WebDrive ia able to more accurately"

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 107ce527-7b9c-ff4e-e524-a8ebbc1a6714
* Version Independent ID: 6c8de68a-01bf-3426-8427-6daa5dca5f03
* Content: [WebDriver (Chromium) - Microsoft Edge Development](https://docs.microsoft.com/en-us/microsoft-edge/webdriver-chromium?tabs=c-sharp&source=docs)
* Content Source: [microsoft-edge/webdriver-chromium.md](https://github.com/MicrosoftDocs/edge-developer/blob/master/microsoft-edge/webdriver-chromium.md)
* Product: **microsoft-edge**
* Technology: **devtools**
* GitHub Login: @MSEdgeTeam
* Microsoft Alias: **msedgedevrel**

jm-trd-ms:
Your feedback resulted in changes to the documentation.
#687

PR #687, opened 2020-05-18
Updated webdriver-chromium

jm-trd-ms:
Updated webdriver-chromium file per customer feedback in Issue #665.


jm-trd-ms:
#sign-off and merged.

PR #690, opened 2020-05-19
Added JavaScript samples since Selenium Tools now available via npm

zoherghadyali:
Preview link here: https://review.docs.microsoft.com/en-us/microsoft-edge/webdriver-chromium?branch=user%2Fzoghadya%2Fselenium-tools-js-update&tabs=javascript#use-microsoft-edge-chromium-with-webdriver

This PR adds JavaScipt code snippets to our WebDriver (Chromium) doc since the Selenium Tools for Microsoft Edge are now available for JavaScript via npm. The package is available here: https://www.npmjs.com/package/@microsoft/edge-selenium-tools and I tested these samples with the following snippet: 

```javascript
const edge = require("@microsoft/edge-selenium-tools");
const {By, Key, until} = require('selenium-webdriver');

(async function example() {
    let options = new edge.Options().setEdgeChromium(true)
    options.setBinaryPath("C:\\Program Files (x86)\\Microsoft\\Edge Dev\\Application\\msedge.exe");
    // options.addArguments("headless");
    // options.addArguments("disable-gpu");
    let service = new edge.ServiceBuilder("C:\\Users\\zoghadya\\Documents\\My_Apps\\DevTools-todomvc\\tests\\edgedriver_win64\\msedgedriver.exe").enableVerboseLogging().build();

    let driver = edge.Driver.createSession(options, service);
    try {
        await driver.get('http://www.google.com/');
        await driver.findElement(By.name('q')).sendKeys('webdriver', Key.RETURN);
        let condition = until.titleIs('webdriver - Google Search');
        await driver.wait(async driver => condition.fn(driver), 10000, 'Loading failed.');
    }
    finally {
        await driver.quit();
        await service.kill();
    }
})();
```

### DIFF
--- a/microsoft-edge/webdriver-chromium.md
+++ b/microsoft-edge/webdriver-chromium.md
@@ -3,7 +3,7 @@ description: Learn how to test your website or app in Microsoft Edge or automate
 title: WebDriver (Chromium)
 author: MSEdgeTeam
 ms.author: msedgedevrel
-ms.date: 05/18/2020
+ms.date: 05/20/2020
 ms.topic: article
 ms.prod: microsoft-edge
 ms.technology: devtools
@@ -88,7 +88,7 @@ var driver = new EdgeDriver();
 driver = Edge()
 ```  
 
-#### [JavaScript](#tab/javascript/)
+#### [JavaScript](#tab/javascript/)  
 
 <a id="basic-usage-code" />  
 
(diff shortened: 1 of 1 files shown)


microsoft-edge/webdriver-chromium.md
```diff
@@ -183,6 +214,19 @@ service_args = ['--verbose']
 driver = Edge(service_args = service_args)
 ```  
 
+#### [JavaScript](#tab/javascript/)  
+
+<a id="customizing-microsoft-edge-driver-services-code" />  
+
+When using JavaScript, the `edge` object has a `ServiceBuilder` method that returns a `service` you may optionally pass to the `edge.driver` object, which creates and starts the service for you.
```
bwalderman, line 221: `ServiceBuilder` is a class, not a method, and the choice of `edge` as a namespace is arbitrary and up to the user, not us (i.e. the user happens to name us `edge` when they require() us). Instead, can we say something like?

Use the `ServiceBuilder` class to create and configure a `Service`. You may optionally pass this `Service` object to the `Driver` object which will start and stop the service for you.

microsoft-edge/webdriver-chromium.md
```diff
@@ -183,6 +214,19 @@ service_args = ['--verbose']
 driver = Edge(service_args = service_args)
 ```  
 
+#### [JavaScript](#tab/javascript/)  
+
+<a id="customizing-microsoft-edge-driver-services-code" />  
+
+When using JavaScript, the `edge` object has a `ServiceBuilder` method that returns a `service` you may optionally pass to the `edge.driver` object, which creates and starts the service for you. 
+
+To configure the `EdgeService`, run additional methods within `edge.ServiceBuilder()` before calling the `build()` method and then pass the service as a parameter in the `edge.Driver.createSession()` method.
```
bwalderman, line 223: `EdgeService` isn't a thing. This should just be `Service`. Also please remove the `edge.` prefixes here too.

PR #706, opened 2020-05-22
Updated webview2/index

jm-trd-ms:
Peer request


jm-trd-ms:
#sign-off offline and merged.

PR #707, opened 2020-05-22
Updated privacy-whitepaper

jm-trd-ms:
Un-commented after //build.


jm-trd-ms:
#sign-off and merged.

PR #735, opened 2020-06-04
updated WinForms getting started guide to reflect not supporting the …

pagoe-msft:
….NET Core designer preview


pagoe-msft:
> Please update the ms-date to 06/03/2020.

Can you set it up so I can just commit the change?

PR #741, opened 2020-06-09
Added deprecation note for hosting and older WebView content

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

Issue #733, opened 2020-06-04
Broken link to https://docs.microsoft.com/en-us/microsoft-edge/hosting/webview2/index


priyasgs:
The first paragraph of the article links to https://docs.microsoft.com/en-us/microsoft-edge/hosting/webview2/index but I see a 404 - Page not found on following the link

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ba9dc9d2-795c-9612-3585-408b4621c080
* Version Independent ID: f73fc16d-d842-4507-9fdb-b42856720355
* Content: [Microsoft Edge WebView 2 for WPF apps - Microsoft Edge Development](https://docs.microsoft.com/en-us/microsoft-edge/webview2/gettingstarted/wpf)
* Content Source: [microsoft-edge/webview2/gettingstarted/wpf.md](https://github.com/MicrosoftDocs/edge-developer/blob/master/microsoft-edge/webview2/gettingstarted/wpf.md)
* Product: **microsoft-edge**
* Technology: **webview**
* GitHub Login: @MSEdgeTeam
* Microsoft Alias: **msedgedevrel**

Reezaali:
@priyasgs thanks for reporting this broken link.

jm-trd-ms:
@priyasgs, 
We appreciate you filing an issue with our team.  
You suggestion has resulted in a change to the documentation.  
Please follow PR #746 for further developments.  

Thank you

PR #746, opened 2020-06-11
Update gettingstarted/wpf file

jm-trd-ms:
Updated link in gettingstarted/wpf.md per customer feedback in #733.


jm-trd-ms:
#sign-off and merged.

PR #748, opened 2020-06-11
Remove additional closing tag for window element

alamathe1:
There is an additional closing tag for the `window` element that fails the project to start on Visual Studio. This PR is to fix the same.


Reezaali:
@alamathe1, thanks for your PR.

jm-trd-ms:
#sign-off and merged.

PR #755, opened 2020-06-16
Updated gettingstarted/win32

jm-trd-ms:
Editorial pass for new content.


jm-trd-ms:
#sign-off offline and merged.

PR #764, opened 2020-06-19
Fix multi-line quoted string doesn't meet yaml standard

mingwli:
multi-line quoted string doesn't meet yaml standard, merge into one line


jm-trd-ms:
#sign-off and merged.

PR #765, opened 2020-06-19
Add Appcache deprecation

ericlaw1979:


jm-trd-ms:
#sign-off offline and merged.

PR #840, opened 2020-07-28
Developer tools, speech recognition, AAD sync, IE

lilliankravitz:
Updates to Developer tools
speech recognition
Sync for AAD sync
Consumer IE Mode


jm-trd-ms:
Duplicates PR #845

PR #841, opened 2020-07-29
Usr/jasteph/debugdoc

jasonstephen15:
Created debugging guide for WebView2.


jasonstephen15:
A couple notes about this doc before it is ready to be holistically reviewed.

- @jm-trd-ms I need some assistance moving images out of this file into the upper level media directory, and still having them display properly in VSCode.
- @Reezaali @pagoe-msft Any opinions on whether we should keep the 3 sections of this document in one doc, or rather create 3 separate files under a debugging node?
- @DavidShoemaker  needed for technical review of doc (specifically VSCode section), VSCode teams mentioned VSCode will inherently support new adapter, do we need to add any support or can we stick with new adapter for this portion?

Reezaali:
@jasonstephen15 - how about using the tab control? You can use a separate tab for each tool.

jasonstephen15:
#sign-off

jm-trd-ms:
#sign-off and merged.

PR #861, opened 2020-08-13
Update winui.md

AE-MS:
Fix reversed x64/x86 links for installing .NET 5.0


jm-trd-ms:
#sign-off and merged.

Issue #864, opened 2020-08-15
Getting started PWA - Service Worker 


don:
The documentation in the  [Service Workers section of the page](https://docs.microsoft.com/en-us/microsoft-edge/progressive-web-apps-edgehtml/get-started#service-workers) needs to be updated for the latest code generated by pwabuilder.com.

Add `type=module` when registering `pwabuilder-sw-register.js` in `layout.pug`

    script(src='/pwabuilder-sw-register.js', type="module")

Change the `offlineFallbackPage` in `pwabuilder.sw.js`

    // const offlineFallbackPage = "ToDo-replace-this-name.html";
    const offlineFallbackPage = "offline.html";

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: bcef5936-232e-1ab0-c1c4-77026c79e277
* Version Independent ID: b6ec450d-ead9-ffcf-0d5b-7a408bf0e13d
* Content: [Get started with Progressive Web Apps - Microsoft Edge Development](https://docs.microsoft.com/en-us/microsoft-edge/progressive-web-apps-edgehtml/get-started#service-workers)
* Content Source: [microsoft-edge/progressive-web-apps-edgehtml/get-started.md](https://github.com/MicrosoftDocs/edge-developer/blob/master/microsoft-edge/progressive-web-apps-edgehtml/get-started.md)
* Product: **microsoft-edge**
* Technology: **pwa**
* GitHub Login: @MSEdgeTeam
* Microsoft Alias: **msedgedevrel**

PR #865, opened 2020-08-15
Getting started PWA - Service Worker 

don:
See #864


jm-trd-ms:
#sign-off and merged.

PR #872, opened 2020-08-20
Chromium docs 5032347

jm-trd-ms:
Updated to the former Feedback section.


jm-trd-ms:
#sign-off and merged.

PR #883, opened 2020-08-31
Chromium docs 5110967

jm-trd-ms:
Updated image formatting for all figures.


jm-trd-ms:
#sign-off and merged.

PR #898, opened 2020-09-11
commit from WebFundamentals

jm-trd-ms:
Updates to images for LOC.


jm-trd-ms:
#sign-off and merged.

PR #905, opened 2020-09-15
Updated redirects.

jm-trd-ms:
Updated redirects in .openpublishing.redirection.json.


jm-trd-ms:
#sign-off and merged.

Issue #868, opened 2020-08-19
Broken link


DenysVuika:
[Enter feedback here]

There's a link on this page to the following resource: https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/network/progressive-web-apps. This link is broken and leads to a page that does not exist.

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 42ed05ac-9e2a-0809-47fb-75919b50d4ba
* Version Independent ID: 00e0f6cf-74f8-29fa-3ac0-78291fd5f428
* Content: [Network Analysis Reference - Microsoft Edge Development](https://docs.microsoft.com/en-gb/microsoft-edge/devtools-guide-chromium/network/reference)
* Content Source: [microsoft-edge/devtools-guide-chromium/network/reference.md](https://github.com/MicrosoftDocs/edge-developer/blob/master/microsoft-edge/devtools-guide-chromium/network/reference.md)
* Product: **microsoft-edge**
* Technology: **devtools**
* GitHub Login: @MSEdgeTeam
* Microsoft Alias: **msedgedevrel**

jm-trd-ms:
@DenysVuika 
Thank you for bringing this issue to our attention.  
Your Issue resulted in an update to the documentation.  
Please follow PR #925 for future updates.

PR #925, opened 2020-09-30
Chromium docs 5229192

jm-trd-ms:
Updated link per feedback in Issue #868 .


jm-trd-ms:
#sign-off and merged.

PR #944, opened 2020-10-07
Update user agent string

alireza-rezaee:


alireza-rezaee:
This is for the official Edge version, not the developer.

PR #986, opened 2020-11-04
Update metadata, part 2 (Chromium docs 5365484)

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #996, opened 2020-11-09
Updated experimental-features.md with composited layers section (Chromium docs 5296991)

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #1024, opened 2020-11-23
Added Service Workers content to DevTools (Chromium docs 5378838)

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #1046, opened 2020-12-15
Updated 11/issues-webhints.msft image file (Chromium docs 5487927)

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #1049, opened 2020-12-17
Adds new article on publishing Chromium-based PWA to the Microsoft Store

JudahGabriel:
Cloned the original EdgeHTML (legacy) article about publishing to the Store, but changed the content to describe the new process of getting PWAs into the Store powered by Chromium-based Edge. Acrolinx score of 91.


JudahGabriel:
@Reezaali The two warnings here are due to the duplicate title. The duplicate title is due to an article of the same name for EdgeHTML, where as this article is Chromium-based Edge.

I think this is acceptable. But let me know if you think we should have a different title.

Reezaali:
@JudahGabriel - can we change it to "Add Progressive Web Apps to the Microsoft Store"? Duplicate H1s aren't good for SEO.

jm-trd-ms (approved):

jm-trd-ms:
#sign-off and merged.

PR #1056, opened 2020-12-28
Updated figures in multiple files.

jm-trd-ms:
Updated figures in multiple files.


jm-trd-ms:
#sign-off and merged.

PR #1090, opened 2021-02-03
Adds Welcome and + tab experiments, moves grid out (Chromium docs 5590130)

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #1142, opened 2021-03-15
Updated icon images (Chromium docs 5767182)

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #1146, opened 2021-03-16
Localization update. (Chromium docs 5663155)

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #1164, opened 2021-03-25
Changes for archival

jm-trd-ms:
Changes for archival of edgehtml content.


jm-trd-ms:
#sign-off and merged.

PR #1166, opened 2021-03-25
Blocking the UI-thread

oggy22:


microsoft-edge/webview2/concepts/threading-model.md
```diff
@@ -39,6 +39,27 @@ Some WebView2 events read values set on the related event arguments or start som
 
 For instance, you may use the `NewWindowRequested` event to provide a `CoreWebView2` to connect as a child window when the event handler completes.  But if you need to asynchronously create the `CoreWebView2`, request the `GetDeferral` method on the `NewWindowRequestedEventArgs`.  After you have asynchronously created the `CoreWebView2` and set the `NewWindow` property on the `NewWindowRequestedEventArgs`, request `Complete` on the `Deferral` object be returned using the `GetDeferral` method.  
 
+## Blocking the UI-thread  
+
+Blocking the UI-thread is usually not a good idea. Blocking the UI-thread doesn't only make the UI unresponsive. Since the WebView2 control can be accessed only from the UI-thread it effectively blocks communication with the browser engine. For example this code would hang because the UI-thread is blocked by waiting for the result:
```
david-risney, line 44: It would be good to describe this in terms we've already mentioned (message pump) and not limit this to .NET

```suggestion
The WebView2 relies on the message pump of the UI thread to run event handler callbacks and async method completion callbacks. If you use methods that block the message pump such as Task.Result or WaitForSingleObject then your WebView2 event handlers and async method completion handlers will not run. For example, the following code will hang because Task.Result stops the message pump waiting for ExecuteScriptAsync to complete, but the message pump is blocked so ExecuteScriptAsync will never complete:
```

microsoft-edge/webview2/concepts/threading-model.md
```diff
@@ -39,6 +39,27 @@ Some WebView2 events read values set on the related event arguments or start som
...
+
+```csharp
+private void Button_Click(object sender, EventArgs e)
+{
+    string result = webView2Control.CoreWebView2.ExecuteScriptAsync("'test'").Result;
+    MessageBox.Show(this, result, "Script Result");
+}
+```
+From another side using `async` and `await` doesn't block the UI-thread:
```
david-risney, line 53:
```suggestion
Instead, you can use an asynchronous await mechanism such as `async` and `await` which do not block the message pump or the UI thread:
```

david-risney (changes requested):

jm-trd-ms (approved):
Updated and ready for #sign-off

pagoe-msft (approved):

pagoe-msft:
#sign-off

jm-trd-ms:
#sign-off and merged.

PR #1167, opened 2021-03-25
Update threading-model.md to explicitly note not to block the UI thread

david-risney:


david-risney:
Oops Ogi already has a PR for this.

PR #1174, opened 2021-03-30
Fixes caption on screenshot

hxlnt:
Screenshot was miscaptioned: It should be "DevTools in Chinese."


jm-trd-ms:
Changes completed in #1177.

PR #1177, opened 2021-03-31
Updated localization article (chromium docs 5859618)

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #1193, opened 2021-04-06
Updated redirects

jm-trd-ms:
Updated redirect to support archiving legacy content.


jm-trd-ms:
#sign-off and merged.

PR #1196, opened 2021-04-07
Updated What's New 89 (Chromium docs 5816462)

jm-trd-ms:


jm-trd-ms:
#sign-off and merged.

PR #1216, opened 2021-04-20
Better images, additional documentation

JudahGabriel:
This PR enhances the article so that the images use the right color and styles for highlighting. It adds information about hosted and classic app packages. Finally, it links to PWABuilder's own documentation that dives deeper into the subject matter.


jm-trd-ms:
#sign-off offline and merged.

jm-trd-ms (approved):

PR #1262, opened 2021-05-28
Correct the manifest namespace to match the required value

ericlaw1979:
Should be `<gupdate xmlns='http://www.google.com/update2/response' protocol='2.0'> ` as it is in https://docs.microsoft.com/en-us/deployedge/microsoft-edge-manage-extensions-webstore

### DIFF
--- a/microsoft-edge/extensions-chromium/enterprise/auto-update.md
+++ b/microsoft-edge/extensions-chromium/enterprise/auto-update.md
@@ -62,7 +62,7 @@ The updated manifest returned by the server should be an XML document.  Review t
 
 ```xml
 <?xml version='1.0' encoding='UTF-8'?>
-<gupdate xmlns='http://www.microsoft.com/update2/response' protocol='2.0'>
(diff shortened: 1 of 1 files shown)


ashishnirvana1:
During a troubleshooting session with one customer it was noticed that if the updated manifest xml file contains:

1.	gupdate xmlns='http://www.google.com/update2/response' -> local hosted extension is successfully installed via ExtensionInstallForcelist
2.	gupdate xmlns='http://www.microsoft.com/update2/response' -> local hosted extension is not installed via ExtensionInstallForcelist

Hence this is good to merge!

ashishnirvana1 (approved):
During a troubleshooting session with one customer it was noticed that if the updated manifest xml file contains:

1.	gupdate xmlns='http://www.google.com/update2/response' -> local hosted extension is successfully installed via ExtensionInstallForcelist
2.	gupdate xmlns='http://www.microsoft.com/update2/response' -> local hosted extension is not installed via ExtensionInstallForcelist

Hence this PR is good to merge.

PR #1296, opened 2021-06-16
Update user-agent-guidance.md

vchapel:
Updated with new Feature Detection content.


microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,114 @@
 ---
-description: This article provides documentation on the Microsoft Edge user agent client hints and user agent string
-title: How to detect Microsoft Edge in your website
+description: This article describes how to detect Microsoft Edge data with user agent client hints and user agent string
+title: Detecting Microsoft Edge from your website
 author: MSEdgeTeam
-ms.author: msedgedevrel
-ms.date: 05/19/2021
+ms.author: v-vchapel
+ms.date: 06/11/2021
```
Reezaali, line 6: Need to update the ms.date field.

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
-ms.date: 05/19/2021
+ms.author: v-vchapel
+ms.date: 06/22/2021
 ms.topic: article
 ms.prod: microsoft-edge
-keywords: microsoft edge, compatibility, web platform, user agent string, ua string, ua overrides, user-agent client hints, user agent client hints, ua client hints, ua ch
+keywords: microsoft edge, compatibility, web platform, user agent string, ua string, ua overrides, user-agent client hints, user agent client hints, ua client hints, ua ch, feature detection, browser identification, browser detection, header, https header, verify chromium, detect microsoft edge, detecting microsoft edge
 ---
-# How to detect Microsoft Edge in your website  
+# Detecting Microsoft Edge from your website
```
mikehoffms, line 11: Might want to change the toc.yml entry too

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
-## User-Agent Client Hints  
+## Feature detection
 
-Starting in Microsoft Edge version 90, access browser information in a cleaner, more privacy-preserving way.  
+Microsoft recommends detecting if a feature is supported in your browser. We do not recommend that you use feature detection to detect a specific browser. For more information see [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection].
 
-### User-Agent Client Hints HTTPS Header  
+If you must detect browsers, we recommend using client hints as follows.
+
+## User Agent Client Hints
```
mikehoffms, line 33: global: User-Agent (w/ hyphen) for consistency.  https://www.bing.com/search?q=%22user+agent+client+hints%22

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -93,31 +111,39 @@ If you need to access more detailed information, use the [getHighEntropyValues()
...
     return false;
 }
-```  
+```
+
+Brand names and the associated display order change over time. Avoid hard-coding brand names.
+
+If you cannot use [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection], don't use a hardcoded list of known Chromium-based browsers for verification. Examples of hardcoded browser names include `Microsoft Edge` and `Google Chrome`.  [Feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection] may not be available because a fix for a Chromium bug in later versions must be avoided and the affected browsers are difficult to detect.
+
+## User Agent string
```
mikehoffms, line 140: possibly globally have a hyphen; User-Agent string.  Debatable.  Should be consistent, in any case.  
- Check https://www.bing.com/search?q=%22user%20agent%20string%22
- Check full-text search in repo, it's both ways.  user agent string vs user-agent string

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Oh of course, changed, thanks for letting me know!

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: I wasn't sure about this one myself ... most in the topic had no hyphen so I went with that ... welcome other opinions @Reezaali

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
-*   Reduce code breakage from changes to the User-Agent string.  
-    
-When [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection] isn't applicable and you must use user agent detection, use the following format to detect Microsoft Edge user agent on Windows and Android.  
+This article describes the methods Microsoft Edge supports for retrieving user agent information.
 
-## User-Agent Client Hints  
+## Feature detection
 
-Starting in Microsoft Edge version 90, access browser information in a cleaner, more privacy-preserving way.  
+Microsoft recommends detecting if a feature is supported in your browser. We do not recommend that you use feature detection to detect a specific browser. For more information see [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection].
```
naarcini, line 29: > We do not recommend that you use feature detection to detect a specific browser.

I don't think we need this sentence. Generally, sites will use either feature detection (ex: Does the current browser support WebRTC? If it does, let's use it) or browser detection (ex: Is the current browser Microsoft Edge? If it is, we know that the browser supports WebRTC so let's use that feature). There aren't "legitimate" cases where a site would use feature detection to guess at which browser it's running in.

microsoft-edge/web-platform/user-agent-guidance.md
naarcini: I think either is fine so long as we're consistent

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
 
-### User-Agent Client Hints HTTPS Header  
+If you must detect browsers, we recommend using client hints as follows.
+
+## User Agent Client Hints
+
+Microsoft Edge supports User-Agent Client Hints starting with version 90.
+
+User-Agent Client Hints easily retrieve basic data such as the browser name and whether or not it is mobile. Also, hints are more likely to be supported in the long term as User-Agent string usage declines.
```
naarcini, line 37: > Also, hints are more likely to be supported in the long term as User-Agent string usage declines.

The primary entity trying to drive a decline in User-Agent string usage is Chrome. We're also trying to do the same, but with less users our efforts are less impactful. Our plans are to move people off the User-Agent string and on to Client Hints.

That in mind, I feel like we'd want to assure people that Client Hints are the way of the future and the User-Agent string is on its way out. I'm not sure what the best way to write that would be. How about something like this:

> Also, hints will continue to be supported in the long term whereas browsers are in the process of reducing the information contained in the User-Agent string.

A link to this might also be useful somewhere:
https://www.chromestatus.com/feature/5704553745874944

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
+- Determine if new browser activity is from the expected user.
+- Customize hints or instructions if the user is new to this site.
+
+Do not use User-Agent Client Hints to:
+- Block *unsupported* browsers.
+- Restrict access to features on your site.
+
+### User Agent Client Hints HTTPS header
+
+When a browser requests a web page from a server, it sends an HTTPS client header request. The server replies with a specific set of HTTPS response header data. User-Agent Client Hints automatically return the brand and mobile attributes. It can easily include additional data.
```
naarcini, line 49: > When a browser requests a web page from a server, it sends an HTTPS client header request

Since we're specifically talking about Edge here, consider something like this phrasing:

> When Microsoft Edge sends an HTTPS request to a server, it sends the User-Agent Client Hints headers that correspond to browser brand and mobile information.

naarcini, line 49: > The server replies with a specific set of HTTPS response header data. User-Agent Client Hints automatically return the brand and mobile attributes. It can easily include additional data.

The server only really needs to do anything if it wants more than the basic (brand and mobile) data. That in mind, consider something like this:

> If the server requires more granular information about the browser, its response includes an `Accept-CH` header. The value of that response header is a comma separated list of all the Client Hints request headers the server wants from the browser (ex: `Accept-CH: Sec-CH-UA-Platform,Sec-CH-UA-Platform-Version`). Microsoft Edge's next HTTPS request to that server will include the requested User-Agent Client Hints headers.

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
 
 By default, Chromium browsers including Microsoft Edge send the `Accept-CH-UA` response header in the following format.
```
naarcini, line 51: Whoops, looks like we had a typo before any of these edits! This should read:

> By default, Chromium browsers including Microsoft Edge send the `Sec-CH-UA` and `Sec-CH-UA-Mobile` request headers in the following format.

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
-Sec-CH-UA: "Chromium";v="91", "Microsoft Edge";v="91","Dummy;Browser Brand";v="99"
+Sec-CH-UA: "Chromium";v="92", "Microsoft Edge";v="92","Placeholder;Browser Brand";v="99"
 Sec-CH-UA-Mobile: ?0
 ```  
 
-> [!NOTE]
-> Client Hints are only sent over secure connections using `HTTPS`.  
-
-The following low entropy hints are sent by default.  If you need to access more information, send one of the following request headers.  
+The following table shows the available header request hints with sample responses. For example, to discover information such as operating system, include `Sec-CH-UA-Platform` in the header request.
```
naarcini, line 58: > The following table shows the available header request hints with sample responses. For example, to discover information such as operating system, include `Sec-CH-UA-Platform` in the header request.

I think this should read simply:

> The following table shows all available hints request headers with sample values.

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
-#### User-Agent request and response headers  
-
-| Request header | Example response value |  
+| User-Agent Request header | Example User-Agent response value |  
 |:--- |:--- |  
 | `Sec-CH-UA` | `"Chromium";v="91", "Microsoft Edge";v="91","GREASE";v="99"` |  
 | `Sec-CH-UA-Mobile` | `false` |  
 | `Sec-CH-UA-Full-Version` | `91.0.866.0` |  
 | `Sec-CH-UA-Platform` | `Windows` |  
 | `Sec-CH-UA Platform-Version` | `10.0` |
```
naarcini, line 67: Another typo we had missed. Could you update this?

> Sec-CH-UA-Platform-Version

microsoft-edge/web-platform/user-agent-guidance.md
naarcini, line 49: I see "User-Agent Client Hints" in here. I'm not sure whether you'll end up choosing to include the hyphen or not (either is fine with me), but let's be consistent. The heading for this section, for example, is not hyphenated

microsoft-edge/web-platform/user-agent-guidance.md
naarcini, line 33: Somewhere early in this section, I think it might be useful to include a link to the spec:
https://wicg.github.io/ua-client-hints/

I know you link it further down, but I feel like an early mention could be helpful

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
-The default response value from `navigator.userAgentData` uses the following format.  
+You can access User-Agent Client Hints using Javascript on the client side. When you call the default `navigator.userAgentData` it returns the following response.  
 
 ```javascript
 { brands: [ {brand: "Chromium","version":"91"}, {brand: "Microsoft Edge","version":"91"}, {brand: "GREASE","version":"99"}, ]
 mobile: false }
 ```  
 
-If you need to access more detailed information, use the [getHighEntropyValues()][GithubWicgUaClientHintsGethighentropyvalues] method.  
+To help you discover issues such as hardcoded brand names, Microsoft Edge includes a `GREASE` brand value that changes over time and is formatted to break if it detects common string parsing issues.
```
naarcini, line 83: > Microsoft Edge includes a `GREASE` brand value that changes over time and is formatted to break if it detects common string parsing issues.

We're not able to detect string parsing issues. The purpose of a changing GREASE value is to break other people's potentially "incorrect" usage of Client Hints. For example, we don't want people to detect Edge version 91 by checking if the brands list is exactly:
`{brand: "Chromium","version":"91"}, {brand: "Microsoft Edge","version":"91"}, {brand: "GREASE","version":"99"}`
Rather, we want them to detect Edge version 91 if and only if there is a brand in that list that is "Microsoft Edge" and the corresponding version is "91", meaning something like this would pass an Edge 91 check too:
`{brand: "Microsoft Edge","version":"91"}, {brand: "Chromium","version":"91"}, {brand: "BROWSER","version":"100"}`

So, I think this should read more like:

> Microsoft Edge includes a `GREASE` brand value that changes over time. Its purpose is to prevent sites from attempting to match the entire brand list when attempting to detect a version of Edge.

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -93,31 +111,39 @@ If you need to access more detailed information, use the [getHighEntropyValues()
...
+For more information navigate to [getHighEntropyValues()][GithubWicgUaClientHintsGethighentropyvalues].
 
-If you're prevented from using [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection], don't use a hardcoded list of known Chromium-based browsers for verification.  Examples of hardcoded browser names include `Microsoft Edge` and `Google Chrome`.  Reasons why [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection] isn't applicable may include the following situation.  
+### Using feature detection with User Agent Client Hints
 
-*   A fix for a Chromium bug in a later version must be avoided and the affected browsers are difficult to detect outside of a verification of the brand and version.  
-    
-Instead, you should verify the `Chromium` brand, which ensures your detection applies to all affected Chromium-based browsers.  
+Combining User-Agent Client Hints with [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection] is the best mechanism for browser identification. Microsoft recommends using it whenever possible to:
```
naarcini, line 118: I mentioned earlier in the review that feature detection is not quite the same thing as browser detection. Using both things is generally good web design, but I don't know if we need to call that out explicitly. Should we omit this section?

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -146,42 +172,42 @@ The response value from `navigator.userAgent` method uses the following format.
...
+The Microsoft Edge Canary and Dev channels don't currently receive user agent overrides. The Microsoft Edge Canary and Dev channels provide environments that use the default Microsoft Edge user agent. Use the Microsoft Edge Canary and Dev channels to reproduce issues on your website caused by the default Microsoft Edge user agent. To turn off user agent overrides in the Microsoft Edge Beta or Stable channels, complete the following actions.  
 
-1.  Open the Microsoft Edge Beta or Stable channel.  
-1.  Navigate to `edge://compat/useragent`.  
-    
-The Microsoft Edge Canary and Dev channels don't currently receive user agent overrides.  The Microsoft Edge Canary and Dev channels provide environments that use the default Microsoft Edge user agent.  Use the Microsoft Edge Canary and Dev channels to easily reproduce issues on your website caused by the default Microsoft Edge user agent.  To turn off user agent overrides in the Microsoft Edge Beta or Stable channels, complete the following actions.  
+1. Open a command prompt. For example, enter **CMD** in the Windows search text box and select the **Command Prompt** app.
+1. Copy and paste the following code snippet.
```
naarcini, line 195: One more typo from before your edits: I think we just want people to copy the snippet below. They paste it in the next numbered step

naarcini (changes requested):
This flows much more nicely than before! We just need to fix up some details

microsoft-edge/web-platform/user-agent-guidance.md
mikehoffms: Microsoft Writing Style Guide in [Hyphens](https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=28765) article says "Hyphenate two or more words that precede and modify a noun as a unit if: ... Confusion might result without the hyphen."  

The below spec hyphenates "User-Agent" and uses init-caps; presents it as a formal technical term.  That's a pretty strong indication to follow suit.  Affects What's New 93 too.

mikehoffms: @Reezaali - pls see spec https://wicg.github.io/ua-client-hints/ which uses "User-Agent", and advise on hyphen & init-caps.

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
 
-Browsers provide mechanisms for websites to detect browser information such as brand and version number.  The mechanisms also detect other device characteristics like the host operating system.  Two such mechanisms supported on Microsoft Edge are [User-Agent Client Hints](#user-agent-client-hints) and [User-Agent strings](#user-agent-string).  After decades of use, User-Agent strings are outdated and have a long history as the cause of website compatibility issues.  Instead, the Microsoft Edge team recommends you use an improved mechanism to retrieve browser information named [User-Agent Client Hints](#user-agent-client-hints).  This article outlines the two mechanisms Microsoft Edge supports for retrieving user agent information.  
+Detecting and restricting access to your site is not recommended because users may not be able to reach your content or products. However, in some cases you may want to provide different experiences to users of different browsers. For example, if you include steps on how to configure Edge or another browser for use with your site, you may want to detect the browser, and then show the appropriate content to that user.
```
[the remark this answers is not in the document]
jenny-microsoft: Hmm I liked that the original gave an overview of what type of information websites can detect via user agent data, as opposed to jumping straight into "don't detect Edge to block it", which should be the use case for a minority of developers. Can we also include a quick sentence in the beginning to outline that?

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
-|  | Server-side | Client-side |  
-|:--- |:--- |:--- | 
-| **User-Agent Client Hints** \(recommended\) | `Sec-CH-UA` HTTPS header | `navigator.userAgentData` JavaScript method |  
-| **User-Agent string** \(legacy\) | `User-Agent` HTTPS header | `navigator.userAgent` JavaScript method |  
+Mechanisms for browser detection with code location and identifiers:
 
-Microsoft recommends that you use [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection] whenever possible for the following reasons.  
+| Mechanism | Code Location |  
+|:--- |:--- | 
+| **Feature Detection** \(recommended\) | Client-side: `navigator.userAgentData.brands` JavaScript method |
```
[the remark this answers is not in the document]
jenny-microsoft: +1. I'd also suggest leaving the comparison for CH vs string so current devs using the UA string can see at a glance what is changing.

microsoft-edge/web-platform/user-agent-guidance.md
jenny-microsoft: Maybe something along the lines of: 

User-Agent Client Hints is a cleaner, more privacy-preserving way to access various information about the browser such as browser name, version number, platform, and more. Over time, the User-Agent string will be frozen and deprecated by various browsers, for example https://www.chromestatus.com/feature/5704553745874944.

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Yes ... I started removing the hyphens in heading as bad form ... but it is better to be consistent. Most sites we reference (and our own) prefer hyphens for both UA stings and UA CH, so I went that direction.

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: fixed

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: I have no idea ... maybe you and @jenny-microsoft and @Reezaali can reach a consensus on this one ...

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: fixed

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Done

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Made the change, thanks for the great example wording

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Removed

microsoft-edge/web-platform/user-agent-guidance.md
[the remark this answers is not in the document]
vchapel: Let me know if the next version is better.

microsoft-edge/web-platform/user-agent-guidance.md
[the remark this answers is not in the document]
vchapel: Tried to include both suggestions in the new version ... let me know, thanks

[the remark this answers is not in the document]
naarcini: I like the way this reads now, thanks!

microsoft-edge/web-platform/user-agent-guidance.md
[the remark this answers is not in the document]
naarcini: The lead-in is definitely better IMO, but I still think the line about Feature Detection doesn't really have a place here. Can we remove that entry from the table?

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -26,15 +28,15 @@ This article describes the methods Microsoft Edge supports for retrieving user a
 
 ## Feature detection
 
-Microsoft recommends detecting if a feature is supported in your browser. We do not recommend that you use feature detection to detect a specific browser. For more information see [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection].
+Microsoft recommends detecting if a feature is supported in your browser. For more information, see [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection].
```
naarcini, line 31: Could we update this to say that we prefer feature detection over any form of browser detection whenever possible?

Ex:

> Microsoft recommends detecting if a feature is supported in your browser whenever possible instead of detecting the browser.

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -26,15 +28,15 @@ This article describes the methods Microsoft Edge supports for retrieving user a
...
-If you must detect browsers, we recommend using client hints as follows.
+If you must detect browsers, Microsoft recommends using client hints as follows.
 
-## User Agent Client Hints
+## User-Agent Client Hints
 
 Microsoft Edge supports User-Agent Client Hints starting with version 90.
 
-User-Agent Client Hints easily retrieve basic data such as the browser name and whether or not it is mobile. Also, hints are more likely to be supported in the long term as User-Agent string usage declines.
+User-Agent Client Hints is a cleaner, more privacy-preserving way to access browser information such as browser name, version number, platform, and more. Soon the User-Agent string will be frozen and deprecated by various browsers. For example, the Chrome Platform Status site describes the change in [Feature: Reduce User-Agent string information][https://www.chromestatus.com/feature/5704553745874944].
```
naarcini, line 39: Looks like the link at the end of the line wasn't quite formatted correctly. When linking to a URL, I think you need parentheses:

Ex: `[Google](https://google.com)` -> [Google](https://google.com)

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -44,9 +46,11 @@ Do not use User-Agent Client Hints to:
 - Block *unsupported* browsers.
 - Restrict access to features on your site.
 
-### User Agent Client Hints HTTPS header
+For more information, navigate to the W3C Community Draft Report at [User-Agent Client Hints][https://wicg.github.io/ua-client-hints/].
```
naarcini, line 49: Same here, looks like we have to fix up this link

microsoft-edge/web-platform/user-agent-guidance.md
naarcini: Were you going to make this change? You could keep "browser" instead of "Microsoft Edge" if you'd like, but I think the rest of my suggestion is a bit more accurate.

naarcini: Were you going to make a change like this?

microsoft-edge/web-platform/user-agent-guidance.md
naarcini: Were you going to make this change?

microsoft-edge/web-platform/user-agent-guidance.md
naarcini: Were you going to make this change?

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Tried to standardize on User-Agent

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Settled on User-Agent

microsoft-edge/web-platform/user-agent-guidance.md
naarcini: Alright, how's this:

- We change the title of this section to something like "User-Agent Client Hints Best Practices" or "User-Agent Client Hints Tips"
- We rephrase this particular part to read:

> Combining User-Agent Client Hints with feature detection is an effective way to deliver compatible web content. Microsoft recommends using this pattern to:

- After the bulleted list, we add a sentence before "Use this method to verify.." that reads something like this (to clarify the code example):

> In addition, Microsoft recommends detecting Chromium, the engine that powers both Microsoft Edge and Google Chrome, when it is necessary to check for a Chrome-like browser.

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -93,31 +115,39 @@ If you need to access more detailed information, use the [getHighEntropyValues()
...
+        if (brand_version_pair.brand == "Chromium"){
             return true;
         }
     }
     return false;
 }
-```  
+```
+
+Brand names and the associated display order change over time. Avoid hard-coding brand names.
```
naarcini, line 140: I think the Brand names will stay constant, but the order in which they're displayed is what will change over time. Can we clarify that here?

naarcini (changes requested):
Thanks for the quick turnaround! I have a few more comments/reminders in this version

microsoft-edge/web-platform/user-agent-guidance.md
mikehoffms: Option 1: A regular link syntax is [link text](URL).  
Option 2: A "reference" link syntax is two lines:
[link text][reference string describing the path]
[reference string describing the path]: url "title of destination | site or type of site"
See examples in repo.

microsoft-edge/web-platform/user-agent-guidance.md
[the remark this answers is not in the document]
jenny-microsoft: Can we keep the old table comparing Client Hints and the User Agent string? Feature detection shouldn't be included here.

microsoft-edge/web-platform/user-agent-guidance.md
jenny-microsoft: +1. Let's keep references to feature detection to a minimum. The general point we want to make is "if you can use feature detection, don't read this article and don't use user agent for browser detection".

microsoft-edge/web-platform/user-agent-guidance.md
jenny-microsoft: Let's replace with "Display order of brand names changes over time. Avoid hard-coding checks for brands at specific indices"

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: The "review" environment doesn't some resolve links ... and I'm not sure why we use [][] instead of []() ... but both work. Maybe ask Reeza  B^)

mikehoffms: Tip for everyone:  When you're reading docs at a "review" url, eg
https://review.docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/whats-new/2021/07/devtools?branch=user%2Fv-mhoffman%2Fwhatsnew93
and you click a link and get a 404, simply delete "review." from the URL and then Refresh; usually this way you can prove that the link will resolve when at the real site.

mikehoffms: re [][] see my comment below, I show the 2 patterns that work.  There are pros & cons (costs) of the long form and I don't know that this policy of which form to use is cast in stone.

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Missed it somehow, fixed now!

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,92 @@
...
+
+Do not use User-Agent Client Hints to:
+- Block *unsupported* browsers.
+- Restrict access to features on your site.
+
+For more information, navigate to the W3C Community Draft Report at [User-Agent Client Hints][https://wicg.github.io/ua-client-hints/].
+
+### User-Agent Client Hints HTTPS header
+
+When a browser requests a web page from a server, it sends an HTTPS client header request. The server replies with a specific set of HTTPS response header data. User-Agent Client Hints automatically return the brand and mobile attributes. It can easily include more data.
```
mikehoffms, line 53:
```suggestion
When a browser requests a web page from a server, it sends an HTTPS client header request. The server replies with a specific set of HTTPS response header data. User-Agent Client Hints automatically returns the brand and mobile attributes. It can easily include more data.

```

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Done

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Replaced the whole thing with better wording. Thanks for checking!

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Missed that the first time, fixed now, thanks!

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
+- Determine if new browser activity is from the expected user.
+- Customize hints or instructions if the user is new to this site.
 
-Starting in Microsoft Edge version 90, access browser information in a cleaner, more privacy-preserving way.  
+Do not use User-Agent Client Hints to:
+- Block *unsupported* browsers.
+- Restrict access to features on your site.
 
-### User-Agent Client Hints HTTPS Header  
+For more information, navigate to the at [W3C Community Draft Report | User-Agent Client Hints][https://wicg.github.io/ua-client-hints/].
```
naarcini, line 46: I think there's a typo here. I think this should be:

> For more information, navigate to the specification at

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
 
-By default, Chromium browsers including Microsoft Edge send the `Accept-CH-UA` response header in the following format.  
+### User-Agent Client Hints HTTPS header
+
+When Microsoft Edge sends an HTTPS request to a server, it sends the User-Agent Client Hints headers that correspond to browser brand and mobile information. If the server requires more granular information about the browser, its response includes an `Accept-CH` header. The value of that response header is a comma-separated list of all the Client Hints request headers the server wants from the browser, such as `Accept-CH: Sec-CH-UA-Platform,Sec-CH-UA-Platform-Version`. The next Microsoft Edge HTTPS request to the server will include the requested User-Agent Client Hints headers.
+
+By default, Chromium browsers including Microsoft Edge send the `Sec-CH-UA` and `Sec-CH-UA-Mobile` response header in the following format.
```
naarcini, line 52: I learned recently that the Brave browser, another Chromium-based browser, isn't intending to support UA client hints. For that reason, I'm thinking we should stick to talking about Edge only in this part. Ex:

> By default, Microsoft Edge sends the

Also, should we use a colon instead of a period at the end of the sentence?

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
-The following low entropy hints are sent by default.  If you need to access more information, send one of the following request headers.  
-
-#### User-Agent request and response headers  
+The following table shows all available hints request headers with sample values.
 
-| Request header | Example response value |  
+| User-Agent Request header | Example User-Agent response value |  
 |:--- |:--- |  
 | `Sec-CH-UA` | `"Chromium";v="91", "Microsoft Edge";v="91","GREASE";v="99"` |  
 | `Sec-CH-UA-Mobile` | `false` |
```
naarcini, line 64: In UA client hints headers, `false` is represented as `?0`. Could we update this?

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,88 @@
...
-| Request header | Example response value |  
+| User-Agent Request header | Example User-Agent response value |  
 |:--- |:--- |  
 | `Sec-CH-UA` | `"Chromium";v="91", "Microsoft Edge";v="91","GREASE";v="99"` |  
 | `Sec-CH-UA-Mobile` | `false` |  
 | `Sec-CH-UA-Full-Version` | `91.0.866.0` |  
 | `Sec-CH-UA-Platform` | `Windows` |  
-| `Sec-CH-UA Platform-Version` | `10.0` |  
+| `Sec-CH-UA-Platform-Version` | `10.0` |  
 | `Sec-CH-UA-Arch` | `x86` |
```
naarcini, line 68: I learned recently that one more UA CH header was added to the spec since this doc was originally written. Can we include one more line below the Arch header:
```
| `Sec-CH-UA-Bitness` | `64` |
```

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -93,31 +111,41 @@ If you need to access more detailed information, use the [getHighEntropyValues()
...
+
+## User-Agent strings
 
-## User agent string  
+User-Agent strings are outdated and have a long history of causing website compatibility issues.
 
-Wherever possible, Microsoft recommends you minimize your use of Microsoft Edge browser detection logic based on the user agent string.  If you have a good reason to detect the user agent, the Microsoft Edge team recommends you use [User-Agent Client Hints](#user-agent-client-hints) as the primary detection logic.  [User-Agent Client Hints](#user-agent-client-hints) also reduces the required number of parsed strings.  However, for legacy reference, the following format is used for the User-Agent string.  
+Wherever possible, Microsoft recommends minimizing use of Microsoft Edge browser detection logic based on the user-agent string. If you have a good reason to detect the user agent, the Microsoft Edge team recommends using [User-Agent Client Hints](#user-agent-client-hints) as the primary detection logic. [User-Agent Client Hints](#user-agent-client-hints) also reduces the required number of parsed strings.
```
naarcini, line 146: nitpick: It feels like we're saying "user-agent string" a lot here. Maybe we can update this phrase:

> If you have a good reason to detect the user agent,

To be something like:

> If you have a good reason to detect the browser,

naarcini: nitpick: The phrasing at the end:

> also reduces the required number of parsed strings.

reads a bit more accurately if it's something like:

> also reduces the complexity in browser detection code.

I say this because I don't know if it's accurate to say that UA CH involves significantly less parsing in general, but it's definitely simpler to implement.

naarcini (approved):
A few more comments left, but I think we're good to go after you resolve these. Thanks for making this page better!

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -93,31 +111,41 @@ If you need to access more detailed information, use the [getHighEntropyValues()
...
+Combining User-Agent Client Hints with [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection] is an effective way to deliver compatible web content. Microsoft recommends using this pattern to:
+* Improve code maintainability.
+* Reduce code fragility.  
+* Reduce code breakage from changes to the User-Agent String.
 
-*   A fix for a Chromium bug in a later version must be avoided and the affected browsers are difficult to detect outside of a verification of the brand and version.  
-    
-Instead, you should verify the `Chromium` brand, which ensures your detection applies to all affected Chromium-based browsers.  
+If you need to check for a Chromium-based browser, Microsoft recommends detecting `Chromium`; the engine that powers both Microsoft Edge and Google Chrome.
```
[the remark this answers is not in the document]
vchapel: I like it, thanks! Fixed.

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Oh yes 🤢  ... fixed.

microsoft-edge/web-platform/user-agent-guidance.md
mikehoffms: https://docs.microsoft.com/en-us/style-guide/punctuation/colons

vchapel: Use colons sparingly ... I'm using that to justify not changing this one ... but it might be a gray area.  Otherwise fixed.

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Added ?0 and ?1 as examples.

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Done

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: Done and done.

microsoft-edge/web-platform/user-agent-guidance.md
naarcini: In each other entry in the table, though, there is only one example. Could we only have one here too? Since we include `?0` above, maybe `?1` would be a good example here

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,89 @@
...
 
-### User-Agent Client Hints HTTPS Header  
+For more information, navigate to the specification at [W3C Community Draft Report | User-Agent Client Hints][https://wicg.github.io/ua-client-hints/].
 
-By default, Chromium browsers including Microsoft Edge send the `Accept-CH-UA` response header in the following format.  
+### User-Agent Client Hints HTTPS header
+
+When Microsoft Edge sends an HTTPS request to a server, it sends the User-Agent Client Hints headers that correspond to browser brand and mobile information. If the server requires more granular information about the browser, its response includes an `Accept-CH` header. The value of that response header is a comma-separated list of all the Client Hints request headers the server wants from the browser, such as `Accept-CH: Sec-CH-UA-Platform,Sec-CH-UA-Platform-Version`. The next Microsoft Edge HTTPS request to the server will include the requested User-Agent Client Hints headers.
+
+By default, Microsoft Edge send the `Sec-CH-UA` and `Sec-CH-UA-Mobile` response header in the following format.
```
naarcini, line 52: Literally 20 mins ago, I found out that there is now one more so-called "Low Entropy" header that will get sent by default: `Sec-CH-UA-Platform`. Consequently, A few edits are in order:

- Above, the example of `Accept-CH: Sec-CH-UA-Platform,Sec-CH-UA-Platform-Version` should be updated to `Accept-CH: Sec-CH-UA-Full-Version,Sec-CH-UA-Platform-Version`
- Here, we need to include the new header:

> By default, Microsoft Edge sends the `Sec-CH-UA`, `Sec-CH-UA-Mobile`, and `Sec-CH-UA-Platform` request headers in the following format.

- Below, we should add an example for `Sec-CH-UA-Platform` (We can use the same one as in the table we have below)

naarcini (approved):
My teammate made me aware of one more very recent change to the spec that we should account for here

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: sent a mock-up in Teams, let me know.

jenny-microsoft (approved):

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,91 @@
...
 { brands: [ {brand: "Chromium","version":"91"}, {brand: "Microsoft Edge","version":"91"}, {brand: "GREASE","version":"99"}, ]
 mobile: false }
 ```  
 
-If you need to access more detailed information, use the [getHighEntropyValues()][GithubWicgUaClientHintsGethighentropyvalues] method.  
+Microsoft Edge includes a `GREASE` brand value that changes over time. It prevents sites from matching the entire brand list when attempting to detect a version of Microsoft Edge.
+
+To request more detailed information such as `platform`, use the following code.
 
 :::row:::
```
Reezaali, line 90: Can we remove the side-by-side format? Let's stack the code blocks vertically instead?

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -93,31 +114,41 @@ If you need to access more detailed information, use the [getHighEntropyValues()
...
+Combining User-Agent Client Hints with [feature detection][MdnLearnToolsTestingCrossBrowserTestingFeatureDetection] is an effective way to deliver compatible web content. Microsoft recommends using this pattern to:
+* Improve code maintainability.
+* Reduce code fragility.  
+* Reduce code breakage from changes to the User-Agent String.
 
-*   A fix for a Chromium bug in a later version must be avoided and the affected browsers are difficult to detect outside of a verification of the brand and version.  
-    
-Instead, you should verify the `Chromium` brand, which ensures your detection applies to all affected Chromium-based browsers.  
+If you need to check for a Chrome-like browser, Microsoft recommends detecting `Chromium`; the engine that powers both Microsoft Edge and Google Chrome.
```
Reezaali, line 126:
```suggestion
If you need to check for a Chrome-like browser, Microsoft recommends detecting `Chromium`, which is the engine that powers Microsoft Edge.

```

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -93,31 +114,41 @@ If you need to access more detailed information, use the [getHighEntropyValues()
...
+        if (brand_version_pair.brand == "Chromium"){
             return true;
         }
     }
     return false;
 }
-```  
+```
+
+Display orders of brand names change over time. Avoid hard-coding checks for brands at specific indices.
```
Reezaali, line 141:
```suggestion
Use the above method to avoid hard-coding checks for brands at specific indices. Display orders of brand names can change over time. 
```

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -93,31 +114,41 @@ If you need to access more detailed information, use the [getHighEntropyValues()
...
+User-Agent strings are outdated and have a long history of causing website compatibility issues.
 
-## User agent string  
+Wherever possible, Microsoft recommends minimizing use of Microsoft Edge browser detection logic based on the User-Agent String. If you have a good reason to detect the browser, the Microsoft Edge team recommends using [User-Agent Client Hints](#user-agent-client-hints) as the primary detection logic. [User-Agent Client Hints](#user-agent-client-hints) also reduces the complexity of browser detection code.
 
-Wherever possible, Microsoft recommends you minimize your use of Microsoft Edge browser detection logic based on the user agent string.  If you have a good reason to detect the user agent, the Microsoft Edge team recommends you use [User-Agent Client Hints](#user-agent-client-hints) as the primary detection logic.  [User-Agent Client Hints](#user-agent-client-hints) also reduces the required number of parsed strings.  However, for legacy reference, the following format is used for the User-Agent string.  
+For legacy reference, the following format was used for User-Agent string.  
 
 :::row:::
    :::column span="":::
```
Reezaali, line 154: Can we remove the side-by-side format?

Reezaali (commented):
Couple of edits. Please review.

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,91 @@
...
+
+> [!NOTE]
+> User-Agent Client Hints are only sent over secure connections using `HTTPS`.
 
 ### User-Agent Client Hints JavaScript API  
 
-The default response value from `navigator.userAgentData` uses the following format.  
+You can access User-Agent Client Hints using JavaScript on the client side. When you call the default `navigator.userAgentData` it returns the following response.  
 
 ```javascript
```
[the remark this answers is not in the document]
vchapel: changed

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: done

microsoft-edge/web-platform/user-agent-guidance.md
vchapel: done

microsoft-edge/web-platform/user-agent-guidance.md
```diff
@@ -1,70 +1,91 @@
...
 { brands: [ {brand: "Chromium","version":"91"}, {brand: "Microsoft Edge","version":"91"}, {brand: "GREASE","version":"99"}, ]
 mobile: false }
```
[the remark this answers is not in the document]
vchapel: this is the only change i left as-is, the format is http, which is how it is coded in the source file.

naarcini:
👏👏👏

PR #1337, opened 2021-07-16
Per 33884847

vchapel:
User agent client hints changes to DevTools


erdraud (approved):
LGTM but @jenny-microsoft should take a look

jenny-microsoft (approved):
Thanks all, LGTM.

PR #1378, opened 2021-08-07
New article "Style editing for CSS-in-JS frameworks"

mikehoffms:
Not shippable yet.  Needs reviewer input on the direction to take this.
Is the TOC title ok?  
Is the top-of-page title ok?
Omitted the video/image that's in the blog post, ok?
This draft has ## Blog heading.  Above that, is derived from What's New.  Below that, is from the blog post.


microsoft-edge/devtools-guide-chromium/css/css-in-js.md
```diff
@@ -0,0 +1,112 @@
...
+
+## What is CSS-in-JS?
+
+The definition of CSS-in-JS is somewhat vague. In a broad sense, it's an approach for managing CSS code using JavaScript. For example, it could mean that the CSS content is defined using JavaScript and the final CSS output is generated on-the-fly by the app.
+
+In the context of DevTools, _CSS-in-JS_ means that the CSS content is injected into the page using CSSOM APIs.  Regular CSS is injected using `<style>` or `<link>` elements, and it has a static source (such as a DOM node or a network resource).  In contrast, CSS-in-JS often doesn't have a static source.  A special case here is that the content of a `<style>` element can be updated using CSSOM API, causing the source to become out of sync with the actual CSS stylesheet.
+
+If you use any CSS-in-JS library (such as styled-component, Emotion, or JSS), the library might inject styles using CSSOM APIs under the hood depending on the mode of development and the browser.
+
+Let's look at some examples on how you can inject a stylesheet using CSSOM API similar to what CSS-in-JS libraries are doing.
```
vchapel, line 54:
```suggestion
Let's look at some examples on how you can inject a stylesheet using CSSOM API similar to how CSS-in-JS libraries work.
```

vchapel: Ending a sentence with "doing" just sounds weird ... it could just end with "similar to CSS-in-JS libraries." or "in the same manner as CSS-in-JS libraries."  I just made a suggestion to start the conversation.

codepo8:
Language and technical explanation is sound here, but I'm confused about the goal of this article. In essence it says that you can apply CSS to elements in different ways and that DevTools then shows the CSS as it is changed. But that's the same if we manipulated the styles object in JavaScript directly. The original article explains how that was done in devtools, which was the news at the time and it is in there since Chrome 85

That said, we have a new feature in there right now that allows you to [copy styles as JS](https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/whats-new/2021/07/devtools#copy-a-declaration-in-the-styles-pane-for-css-in-js-libraries) which would tie in much nicer to this story and bring it more up-to-date.

mikehoffms:
@codepo8 @zoherghadyali Please re-review this PR regarding the positioning of these capabilities in the Styles pane in the case of css-in-js libraries/frameworks.  I added a section about **Copy declaration as JS**.

PR #1406, opened 2021-08-21
Update code sample text to match the image shown

cjfravel-msft:


mikehoffms (commented):
Thank you for the PR.

**Comments from the Edit/review pass:**
I'm going to Close this PR without Merging it to master branch & live.

-  In the text, 'select' (the existing text) is correct per the "Select" row at the Style Guide https://docs.microsoft.com/en-us/style-guide/procedures-instructions/describing-interactions-with-ui
   - The image is incorrect in using the word 'click'.  Also the image needs to be resized to 50% - I'm informing someone who is doing an image update pass.
- The existing capitalization in the h1 is correct and consistent, given that this doc-set uses (by design) the style convention of sentence case rather than title case.  Were we to use title case, 'day' would also need cap'zn.  Check MWSG & DCG to look for a directive to use title case in top-of-page titles:
MWSG public version - article "Formatting titles" - https://docs.microsoft.com/en-us/style-guide/text-formatting/formatting-titles - "Titles of web articles ... use sentence case".  (I'd argue for title case as being easiest and most helpful for readers, but MWSG says use sentence case.)

PR #1536, opened 2021-10-27
Writer/Editor pass on PWA articles

mikehoffms:
**Rendered articles for review:**
https://review.docs.microsoft.com/en-us/microsoft-edge/progressive-web-apps-chromium/?branch=pr-en-us-1536


microsoft-edge/progressive-web-apps-chromium/index.md
```diff
@@ -119,48 +90,45 @@ With you PWAs, you can:
...
 
-Learn how to [publish your PWA to the Microsoft Store](./how-to/microsoft-store.md).
+Learn how to [publish your PWA to the Microsoft Store](how-to/microsoft-store.md).
 
 
 <!-- ====================================================================== -->
 ## Success stories
 
-Using PWA technologies is a great way to make your app **safe**, **discoverable**, **linkable**, **easy to install** and update, **responsive**, and **network independent**. Many businesses use PWAs and have had success.
+Using Progressive Web App (PWA) technologies is a great way to make your app **safe**, **discoverable**, **linkable**, **easy to install** and update, **responsive**, and **network independent**.  Many businesses have had success with PWAs.  For example:
```
captainbrosset, line 111: We keep using both `Progressive Web App` and `PWA` throughout these articles. And we often put them next to each other: `Progressive Web App (PWA)`. 
Do you think we need to do this? Or could we perhaps use `Progressive Web App` everywhere, and drop `PWA`?
Or, alternatively, use `Progressive Web App (PWA)` on this index page, at the beginning, to help new users understand the term, and then just use `PWA` everywhere else?

microsoft-edge/puppeteer/index.md
```diff
@@ -3,109 +3,115 @@ description: Use Puppeteer to automate and test in Microsoft Edge.
...
 
 <!-- ====================================================================== -->
 ## Launch Microsoft Edge with puppeteer-core
 
-> [!NOTE]
-> `puppeteer-core` relies on Node v8.9.0 or later.  The example below uses `async`/`await` which is only supported in Node v7.6.0 or later.  Run `node -v` from the command-line to ensure you have a compatible version of Node.js.
+`puppeteer-core` is similar to other browser-testing-frameworks, such as [WebDriver](../webdriver-chromium/index.md).  You create an instance of the browser, open a webpage, and then manipulate the webpage by using the Puppeteer API.
 
-`puppeteer-core` should be familiar to users of other browser-testing-frameworks like [WebDriver](../webdriver-chromium/index.md).  You create an instance of the browser, open a page, and then manipulate it with the Puppeteer API.  In the following code sample, `puppeteer-core` launches Microsoft Edge, navigates to `https://www.microsoftedgeinsider.com`, and saves a screenshot as `example.png`.
+Microsoft Edge (EdgeHTML) doesn't work with `puppeteer-core`.  You must install the [Microsoft Edge Insider channels](https://www.microsoftedgeinsider.com/download) to follow this example.
```
captainbrosset, line 40: This is out of scope for this PR, but we should update this sentence. EdgeHTML-versions of Edge are no longer available, and there is no need to go the one of the insiders channels to get Chromium-versions of Edge.

microsoft-edge/puppeteer/index.md
```diff
@@ -3,109 +3,115 @@ description: Use Puppeteer to automate and test in Microsoft Edge.
 title: Puppeteer overview
```
captainbrosset, line 3: Note for the future: it makes it easier to review PRs when they're focused on just one thing. In this case: an editor pass on PWA articles.
Updating this Puppeteer article feels a bit out of scope. It makes reviewing the PR harder for me. I don't have a lot of knowledge at all about Puppeteer, so I don't really feel comfortable signing off on these changes.
That said, you are not changing content here, so it's fine. But I think in the future it would help us to keep PRs as narrowly focused as we can.

microsoft-edge/progressive-web-apps-chromium/how-to/share.md
```diff
@@ -71,31 +71,37 @@ async function shareSomeContent(title, text, url) {
...
+:::image type="content" source="../media/windows-share-dialog.png" alt-text="The share dialog on Windows.":::
 
-:::image type="content" source="../media/windows-share-dialog.png" alt-text="The share dialog on Windows." lightbox="../media/windows-share-dialog.png":::
+Once the user has chosen an app to receive the shared content, it is up to this app to handle it any way it chooses.  For example, an email app might use the `title` as the email subject, and use the `text` as the email body.
 
-Once the user has chosen an app to receive the shared content, it is up to this app to handle it any way it chooses. An email app might use the `title` as the email subject and the `text` as the body, for example.
 
-### Sharing files
+<!-- ====================================================================== -->
+## Sharing files
```
captainbrosset, line 82: I don't think changing the h3 in h2 is a good idea in this article.

There are 2 main directions to sharing: sharing from an app, and accepting shared data from another app.
I believe we want to keep these 2 as parent sections to the rest.

For example, this section `Sharing files` "belongs" to the `Sharing content` h2 section. It only makes sense in this direction (sharing from an app).

For the other way around, the `Register as a target`, `Handle GET shared data`, `Handle POST shared data`, and `Handle shared files` sections belong to the `Receiving shared content` h2 section. They don't make sense on their own.

I see you've added the phrase `The following sections cover receiving shared content.` further down to help with this.
So I imagine you made this change so that the table of content displayed on the page would be more useful. Is that correct?

microsoft-edge/progressive-web-apps-chromium/how-to/shortcuts.md
```diff
@@ -13,24 +13,24 @@ keywords: progressive web apps, PWA, Edge, Windows, Taskbar, jumplist, shortcut
...
-:::row-end:::
+The next figure shows shortcuts on Microsoft Edge on iOS:
 
-PWAs can also define common tasks as shortcuts in the [Web App Manifest](./web-app-manifests.md).
+:::image type="content" source="../media/edge-ios-shortcuts.png" alt-text="Shortcuts on Microsoft Edge on iOS.":::
+
+<!-- remove/replace screenshot? -->
+The next figure shows a jumplist on the Twitter app on Windows:
+
+:::image type="content" source="../media/pwa-shortcuts-in-taskbar.png" alt-text="Jumplist on the Twitter app on Windows.":::
```
captainbrosset, line 25: The screenshot is actually not showing the Twitter app at all. I had replaced it with the Webboard demo app, but forgot to change the alt-text at the time.
So the right thing to do here is change the sentence you added to something like "The next figure shows a jumplist on the Webboard app on Windows:", and change the alt-text accordingly.

captainbrosset (changes requested):

microsoft-edge/progressive-web-apps-chromium/index.md
mikehoffms: "We often put them next to each other":
https://docs.microsoft.com/en-us/style-guide/acronyms
"In general, include the acronym in parentheses following the spelled-out term. On subsequent mentions in the same article, page, or screen, you can use the acronym without spelling it out."

"Could we perhaps use Progressive Web App everywhere, and drop PWA?"
Depends: How often is "PWA" used in the industry?  I'm assuming that the industry heavily uses both "PWA" and Progressive Web Application, so I'm keeping both, as-is, which is: in any article that uses "PWA", define it in the 1st occurrence.  I also define it in the first occurrence in each h2 section - that's debatable.

We cannot assume the word "PWA" means anything to the reader; it's safer to use the spelled-out term, at least at top of every article, and possibly at top of each h2 section, as done in this PR.

On the hypertext & Docs principle of "Every page is Page 1", we must assume the reader of any article hasn't seen the "first" article.  Per MWSG, and per standard Docs practice, an acronym must be defined in every article: "On subsequent mentions in the same article, page, or screen, you can use the acronym without spelling it out."

microsoft-edge/progressive-web-apps-chromium/how-to/index.md
```diff
@@ -138,91 +148,74 @@ Your web app now has a service worker that uses the cache-first strategy.  The n
...
+
+Support [Deep linking](https://en.wikipedia.org/wiki/Deep_linking).  Route each page of your site to a unique URL so existing users can help you engage an even broader audience through social media sharing.
+
+### Provide a rich offline experience
+
+Provide a rich [Offline experience](./offline.md).  Make the app work even if the user's device is offline.  Provide a custom offline page rather than using the browser's default offline page.
+
+### Use validation and testing practices
+
+Use software [Validation and testing practices](https://webhint.io).  Use code quality tools like the [Webhint](https://webhint.io) linter to optimize the efficiency, robustness, safety, and accessibility of your app.
```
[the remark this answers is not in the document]
mikehoffms: Agreed, having dup links close together is confusing; avoid overuse of links.

microsoft-edge/progressive-web-apps-chromium/how-to/index.md
```diff
@@ -138,91 +148,74 @@ Your web app now has a service worker that uses the cache-first strategy.  The n
...
+*   CSS [grid](https://developer.mozilla.org/docs/Web/CSS/CSS_Grid_Layout).
+*   [Flexbox](https://developer.mozilla.org/docs/Web/CSS/CSS_Flexible_Box_Layout).
+*   [Media queries](https://developer.mozilla.org/docs/Web/CSS/Media_Queries).
+*   [Responsive images](https://developer.mozilla.org/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images).
+
+To test your app locally, use [device emulation tools](../../devtools-guide-chromium/device-mode/testing-other-browsers.md) from your browser.  To test your app directly on a target device, create a remote debugging session on [Windows](../../devtools-guide-chromium/remote-debugging/windows.md) or [Android](../../devtools-guide-chromium/remote-debugging/index.md).
+
+### Support deep linking
+
+Support [Deep linking](https://en.wikipedia.org/wiki/Deep_linking).  Route each page of your site to a unique URL so existing users can help you engage an even broader audience through social media sharing.
```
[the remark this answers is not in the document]
mikehoffms: These are inline spaces, not trailing spaces.

Two space characters between sentences makes it easier for me to improve docs during editing, to see where there are separate sentences.  This repo ususally used two spaces, and that's one convention that I liked.  Don't know _why_ it used two spaces.

When I use two space characters, as I have opportunistically done in this repo based on its established use of two spaces, it's for the same reason as two blank spaces above an h2 heading in the .md file: to see the structure more easily.  And same reason as break up paragraphs that are longer than ~3 lines in .md.  AFAIK, most tech writers use a single space; AFAIK, this repo was unusual for using two spaces.  I suspect there's a minority of tech writers who prefers two spaces, to more easily see where sentences end and start.

In Markdown, two _trailing_ spaces, with a non-blank line below it, means manual line break, same as <br/> tag.

microsoft-edge/progressive-web-apps-chromium/how-to/index.md
```diff
@@ -124,11 +132,13 @@ Service workers are specialized [Web Workers](https://developer.mozilla.org/docs
 
 Service workers are defined in a special JavaScript file, described in [Using Service Workers](https://developer.mozilla.org/docs/Web/API/Service_Worker_API/Using_Service_Workers) and [Service Worker API](https://developer.mozilla.org/docs/Web/API/Service_Worker_API).
 
-To build a service worker in your project, use the **Cache-first network** service worker recipe from PWA Builder, as follows.
+To build a service worker in your project, use the **Cache-first network** service-worker recipe from PWA Builder, as follows.
```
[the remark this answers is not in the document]
mikehoffms: "this phrase" could mean "service worker recipe" or just "service worker".  The phrase "service worker recipe", case-sensitive, only occurs one time in the repo: this instance.  The only other variant is: "Service Worker recipes", which shows an alternative solution instead of hyphen: init caps.  I might change this instance to init caps instead of hyphen.

See MWSG https://docs.microsoft.com/en-us/style-guide/punctuation/dashes-hyphens/hyphens#noun-modifiers

microsoft-edge/progressive-web-apps-chromium/how-to/handle-protocols.md
```diff
@@ -11,42 +11,40 @@ keywords: progressive web apps, PWA, Edge, JavaScript, protocol
...
+To create a more engaging experience, Progressive Web Apps can handle multiple protocols, such as `http` or `mailto`.
 
-A URL (or Uniform Resource Locator) is a string that allows clients to identify a resource on the Web, and it contains an initial part, the protocol, that lets clients know how to retrieve the resource.
-
-Examples of protocols include:
+A URL (or Uniform Resource Locator) is a string that allows clients to identify a resource on the Web.  The first part of a URL is the _protocol_, which lets clients know how to retrieve the resource.  Examples of protocols include:
 
 | Protocol | Example |
 |:--- |:--- |
 | `http` | http://contoso.com/ |
```
mikehoffms, line 20: @captainbrosset -  the table has `| `http` | http://contoso.com/ |` - I was hoping you'd address that, the source of the revised phrasing of the sentence above it.

microsoft-edge/progressive-web-apps-chromium/how-to/window-controls-overlay.md
```diff
@@ -119,13 +122,12 @@ You can use the `-webkit-app-region` CSS property to offer more ways for users t
...
 
-You can use the `geometrychange` event and the `visible` property on the `navigator.windowControlsOverlay` object to listen to changes and know if the title bar is visible.
+To listen for changes, use the `geometrychange` event.  To detect whether the title bar is visible, use the `visible` property on the `navigator.windowControlsOverlay` object.
 
 > [!NOTE]
-> The `geometrychange` is fired very frequently when the user resizes the window. To avoid running layout-changing code too often and cause performance problems in your app, it is recommended to use a `debounce` function to limit how many times the event is handled.
-> To learn more about `debounce`, see [The Difference Between Throttling and Debouncing](https://css-tricks.com/the-difference-between-throttling-and-debouncing/).
+> The `geometrychange` is fired very frequently when the user resizes the window.  To avoid running layout-changing code too often and causing performance problems in your app, use a `debounce` function to limit how many times the event is handled.  See [The Difference Between Throttling and Debouncing](https://css-tricks.com/the-difference-between-throttling-and-debouncing/).
 
 ```javascript
```
mikehoffms, line 132: Global for PWA articles: wrap code at 85 per DCG.

microsoft-edge/progressive-web-apps-chromium/how-to/window-controls-overlay.md
```diff
@@ -15,39 +15,40 @@ While it is possible for PWAs on mobile to define how they want to be displayed
 
 By default, the app area starts right below the reserved title bar area.
```
mikehoffms, line 16: change to "desktop PWAs can't use this APPROACH to create an immersive, native-like, experience"

microsoft-edge/progressive-web-apps-chromium/how-to/handle-protocols.md
captainbrosset: This table is used to illustrate the general introduction to what URLs and protocols are from the previous sentence. It's not meant to list protocols that PWAs might decide to handle. 
Because it's used as a general introduction, it makes sense for it to list very well known protocols, and http and mailto happen to be (I think) the 2 most well known protocols.

So, while it makes sense to have a small part about what a protocol even is (and use the http example there), it doesn't make as much sense to say that PWAs might want to handle http.

Maybe we should rephrase the start of the article to something like this:

```
A URL (or Uniform Resource Locator) is a string that allows clients to identify a resource on the Web. The first part of a URL is the _protocol_, which lets clients know how to retrieve the resource. Examples of protocols include http or mailto.

To create a more engaging experience, Progressive Web Apps can handle protocols too.
```

So we skip the table. It's a bit too prominent and might give the wrong idea that it's a reference of what protocols PWAs can handle.

And then we go straight to the rest:

```
With Protocol Handling, when a PWA is being installed, the PWA can register to handler certain allowed protocols. ...
```

microsoft-edge/progressive-web-apps-chromium/how-to/share.md
captainbrosset: As discussed offline, this was done as a way to promote h3s to h2s as a way for them to show up in the TOC.
I feel like it breaks the structure of the document a little bit, but at the same time, is this even really visible to readers? The font-size of h2 and h3 isn't very different.
So I'm good either way.

microsoft-edge/progressive-web-apps-chromium/index.md
captainbrosset: Thanks. Having the "Progressive Web Apps (PWAs)" for the first mention of every page makes a lot of sense. I would tend to use PWA only for subsequent mentions though.

microsoft-edge/progressive-web-apps-chromium/how-to/index.md
```diff
@@ -12,7 +12,9 @@ keywords: progressive web apps, PWA, Edge, Windows, PWABuilder, web manifest, se
 
 <!-- todo: make sure screenshots of DevTools are up-to-date -->
 
-Progressive Web Apps (PWAs) are web apps that are [progressively enhanced](https://en.wikipedia.org/wiki/Progressive_enhancement).  The progressive enhancements include app-like features, such as installation, offline support, and push notifications.  You can also package your PWA for app stores, such as Microsoft Store, Google Play, and Mac App Store.  The Microsoft Store is the commercial app store that's built into Windows 10.
+Progressive Web Apps (PWAs) are web apps that are [progressively enhanced](https://en.wikipedia.org/wiki/Progressive_enhancement).  The progressive enhancements include app-like features, such as installation, offline support, and push notifications.
+
+You can also package your PWA for app stores, such as Microsoft Store, Google Play, and Mac App Store.  The Microsoft Store is the commercial app store that's built into Windows 10.
```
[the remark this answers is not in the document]
captainbrosset: Unresolving this since this was not addressed. Is there a reason to keep Windows 10 here?

captainbrosset (changes requested):

microsoft-edge/progressive-web-apps-chromium/how-to/share.md
mikehoffms: h2s were changed back to h3s, in latest .md file diff (not visible in Outdated diff).

microsoft-edge/progressive-web-apps-chromium/how-to/index.md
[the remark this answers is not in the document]
mikehoffms: This was, in fact, fixed.  See the latest .md diff view, not this Outdated view.  Not sure the most efficient way to navigate to latest .md diff view.  The text now reads "built into Windows 10 and later."

captainbrosset (approved):

PR #1564, opened 2021-11-10
Streamlined intro of "About Panel and Drawer tools"

mikehoffms:
**Rendered article for review:**
https://review.docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/about-panel-drawer-tools?branch=pr-en-us-1564


Reezaali (approved):
Suggestion - can we change the H2s to a 2-column table, where the 1st column is the task to do, and the 2nd column are the steps to perform?

PR #1572, opened 2021-11-17
Automatically label new issues with the 'triage-needed' label

captainbrosset:
This introduces a new workflow that adds the `triage-needed` label to new (and re-opened) issues.
This way, we can easily filter issues with this label and go through a regular triage meeting to handle requests to change/fix docs.

I tested this workflow on my forked version of the repo:

* here is an example of an issue I opened: https://github.com/captainbrosset/edge-developer/issues/7
* here is the evidence that the workflow did run successfully: https://github.com/captainbrosset/edge-developer/actions/runs/1471041515

### DIFF
--- a/.github/workflows/label-new-issues.yaml
+++ b/.github/workflows/label-new-issues.yaml
@@ -1,4 +1,4 @@
-# This workflow is used to add the 'triage' label to newly opened issues.
+# This workflow is used to add the 'triage-needed' label to newly opened issues.
 
 name: Label new issues
 on:
@@ -15,5 +15,5 @@ jobs:
       - name: Label issues
         uses: andymckay/labeler@5c59dabdfd4dd5bd9c6e6d255b01b9d764af4414
         with:
-          add-labels: "triage"
+          add-labels: "triage-needed"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/label-new-issues.yaml
+++ b/.github/workflows/label-new-issues.yaml
@@ -1,0 +1,19 @@
+# This workflow is used to add the 'triage' label to newly opened issues.
+
+name: Label new issues
+on:
+  issues:
+    types:
+      - reopened
+      - opened
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/label-new-issues.yaml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  label_issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Label issues
+        uses: andymckay/labeler@5c59dabdfd4dd5bd9c6e6d255b01b9d764af4414
+        with:
+          add-labels: "triage-needed"
```
Reezaali, line 18: I see triage added to new issues, not triage-needed. Bug?

.github/workflows/label-new-issues.yaml
```diff
@@ -0,0 +1,19 @@
+# This workflow is used to add the 'triage-needed' label to newly opened issues.
```
Reezaali, line 1: should this file be moved to the scripts folder?

Reezaali (commented):
Left some comments. I also tested the creation of an issue which looks like it ran successfully but as I noted below, I'm confused about the label that was added. 

https://github.com/captainbrosset/edge-developer/issues/8

.github/workflows/label-new-issues.yaml
captainbrosset: No it's just that when I tested on my fork, I used the `triage` label. But I later saw that this repository already had the `triage-needed` label. So I thought it'd be better to use it instead of `triage`. I just didn't run the test again on my fork.

.github/workflows/label-new-issues.yaml
captainbrosset: No github workflows have to be in this specific directory. We can't do anything about it.

Reezaali (approved):

Reezaali:
@mikehoffms - you can proceed to merge when ready.

PR #1606, opened 2021-12-01
WebView2 ToC change - Merge "Concepts" and "How to" nodes and sort titles alphabetically

Reezaali:
**Rendered TOC for review:**
https://review.docs.microsoft.com/en-us/microsoft-edge/webview2/how-to/webdriver?branch=pr-en-us-1606

### DIFF
--- a/microsoft-edge/toc.yml
+++ b/microsoft-edge/toc.yml
@@ -459,12 +459,22 @@
               href: webview2/get-started/wpf.md
         - name: Sample code
           href: webview2/code-samples-links.md
-        - name: Concepts
+        - name: Concepts and How-To
           items:
+            - name: Automate with Edge Driver
+              href: webview2/how-to/webdriver.md
+            - name: Debug WebView2
+              href: webview2/how-to/debug.md
+            - name: Development best practices
+              href: webview2/concepts/developer-guide.md
             - name: Differences between Microsoft Edge and WebView2
               href: webview2/concepts/browser-features.md
             - name: Distribute a WebView2 app and Runtime
               href: webview2/concepts/distribution.md
+            - name: Embed web content into native applications
+              href: webview2/how-to/communicate-btwn-web-native.md
+            - name: Extended JavaScript scenarios in WebView2
+              href: webview2/how-to/javascript.md
             - name: Manage the user data folder
               href: webview2/concepts/user-data-folder.md
             - name: Managing WebView2 applications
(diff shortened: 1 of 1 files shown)


microsoft-edge/toc.yml
```diff
@@ -459,12 +459,22 @@
               href: webview2/get-started/wpf.md
         - name: Sample code
           href: webview2/code-samples-links.md
-        - name: Concepts
+        - name: Concepts and How-To
```
mikehoffms, line 462:
```suggestion
        - name: Concepts and How-to
```
Or, we could go with just "How to", like done in PWA: https://review.docs.microsoft.com/en-us/microsoft-edge/progressive-web-apps-chromium/how-to/?branch=pr-en-us-1606
The defined Content Types are designed for Azure services docs (using highly structured templates & TOC parent node labels for them), rather than for product docs, so there's only loose guidance (if any) from Docs Contributor Guide, for this TOC bucket label for this repo.

microsoft-edge/toc.yml
```diff
@@ -459,12 +459,22 @@
...
           items:
+            - name: Automate with Edge Driver
```
mikehoffms, line 464: I don't think alphabetical ordering is correct; it's random ordering.  I think @jasonstephen15 should arrange these nodes in a flat, logically grouped sequence to put related articles close together.

Note about nesting/depth: If we discover a group of 4 highly related articles, _possibly_ create a topical parent node to group them - though I'm told that nesting should be avoided, and it's actually good (for accessibility & usability) to have more than 7 siblings in a node (possibly many more).

Reezaali: We'll address grouping in v2.

microsoft-edge/toc.yml
```diff
@@ -459,12 +459,22 @@
               href: webview2/get-started/wpf.md
         - name: Sample code
           href: webview2/code-samples-links.md
-        - name: Concepts
+        - name: How to
```
champnic, line 462: Can we call the combined section "Concepts and How To"?

microsoft-edge/toc.yml
```diff
@@ -459,12 +459,22 @@
...
           items:
+            - name: Automate with Edge Driver
+              href: webview2/how-to/webdriver.md
+            - name: Debug WebView2
+              href: webview2/how-to/debug.md
+            - name: Development best practices
```
champnic, line 468: To avoid unnecessary change, can we just have the two lists appended, rather than resorted? In the short term I think it will just be confusing to have docs moving around, especially if we are planning more moves or grouping in the near future.

champnic (changes requested):

mikehoffms (approved):

jasonstephen15 (approved):

microsoft-edge/toc.yml
mikehoffms: Finding: Alphabetical TOC is used in these special scenarios:
* Auto-gen'd API Ref.  https://docs.microsoft.com/en-us/microsoft-edge/webview2/webview2-api-reference  User is likely to look up in alpha order.
* List of tools, which matches the order in the Command Menu (separate groups for Panel & Drawer tools) & More Tools menu (interleaved Panel & Drawer tools).  https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/about-panel-drawer-tools

Reezaali: Added work item to the backlog to address grouping. To reduce confusion, I'll proceed to sort alphabetically in the meantime.

PR #1621, opened 2021-12-08
Updated demo links and screenshots in Workspaces & Network tutorials

captainbrosset:
I just moved 2 of our tutorial demos from glitch to github (we have a new demos repository [here](https://github.com/MicrosoftEdge/Demos)).

In this PR, I'm making the corresponding updates to the tutorials. This PR contains 2 changes:

1. Change to the DevTools workspaces tutorial.
    In this one I changed the demo URL, removed a screenshot of glitch that was no longer needed, and removed a comment that mentioned filing issues on the alphabet's forum (we're never going to use this comment, so this felt like a quick and simple change to add to this PR).

   **Before**: https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/workspaces/
   **After**: https://review.docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/workspaces/index?branch=pr-en-us-1621

1. Change to the network tutorial demo app moved to github.
    In this one, I changed the demo URL too, but I also updated all the screenshots since they showed information about the old demo that had changed (domain names and URLs). While doing this, I also took the liberty to crop the screenshots to just the devtools portion (where possible). They used to show the web content, for no good reason.

   **Before**: https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/network/
   **After**: https://review.docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/network/index?branch=pr-en-us-1621

There are many other improvements we could make to these tutorials, but I didn't make them to keep the PR focused on mostly the demo URL changes.


mikehoffms (approved):

mikehoffms:
Minor FYI note: both versions of Docs Contributor Guide say to avoid Alerts.  https://docs.microsoft.com/en-us/contribute/markdown-reference#alerts-note-tip-important-caution-warning - "Avoid Notes, Tips, and Important boxes. Readers tend to skip over them. It's better to put that info directly into the article text.  If you need to use alerts, limit them to one or two per article. Multiple alerts should never be next to each other in an article."

PR #1635, opened 2021-12-15
Changed the description to be compatible with Chromium Edge.

tatsuya-midorikawa:
Changed the description from using classes for Legacy Edge  to using classes for Chromium Edge.


Reezaali:
Hi @tatsuya-midorikawa - thanks for the contribution. I spoke to Engineering and we plan to make some updates to the docs by next week. I'll go ahead and close this PR for now because of those incoming updates. 
Thanks again for the contribution!

PR #1641, opened 2021-12-16
Added article "Analyze HTML pages using the Inspect tool"

codepo8:
An overview article on how to use the Inspect tool, including the trick to press Shift to select hidden elements. This was a request from Twitter.

**Rendered article for review:**
https://review.docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/css/inspect?branch=pr-en-us-1641


microsoft-edge/devtools-guide-chromium/css/inspect.md
```diff
@@ -0,0 +1,82 @@
...
+ms.topic: article
+ms.prod: microsoft-edge
+keywords: microsoft edge, web development, f12 tools, devtools
+---
+
+# Analyze HTML pages using the Inspect tool 
+
+This article shows you how to use the **Inspect** tool to preview information about and select an element from the current document. You can try it out yourself by checking the [Inspect Demo Page](https://microsoftedge.github.io/DevToolsSamples/inspector/inspector-demo.html).
+
+The **Inspect** tool button (![Inspect](../media/inspect-icon.msft.png)) is in the upper-left corner of DevTools.  When you select the **Inspect** tool button, the button turns blue, indicating that the **Inspect** tool is active.
```
captainbrosset, line 16: It looks like you reused an existing icon from another page. The problem with this is that the icon in your next screenshot is different from this one, and that may be confusing.

I know it feels like having a shared library of icons would be better, but we actually prefer to have images be per page instead. Even if that means having several copies of the same image.
Could you please create a new screenshot of the inspect icon just for this article?

The other related thing here is: we normally use the default theme for all screenshots. Do you think it'd be ok to take those screenshots again but using the Light+ theme instead?

microsoft-edge/devtools-guide-chromium/css/inspect.md
```diff
@@ -0,0 +1,82 @@
...
+
+:::image type="complex" source="images/inspect-tool-button.msft.png" alt-text="The inspect tool button on the top right of the Developer Tools" lightbox="images/inspect-tool-button.msft.png":::
+   The inspect tool button on the top right of the Developer Tools
```
captainbrosset, line 19: This says "top right", but it should say "top left".

microsoft-edge/devtools-guide-chromium/css/inspect.md
```diff
@@ -0,0 +1,82 @@
...
+:::image-end:::
+
```
captainbrosset, line 21: No need for 2 empty lines between the image and the line below. 1 is enough here.

microsoft-edge/devtools-guide-chromium/css/inspect.md
```diff
@@ -0,0 +1,82 @@
...
+If the element has been positioned using CSS grid or CSS flexbox you get different icons displayed next to its name.
+
+:::image type="complex" source="images/inspect-tool-grid-element.msft.png" alt-text="An element displayed using CSS grid layout showing an extra icon next to its name in the Inspect overlay" lightbox="images/inspect-tool-grid-element.msft.png":::
+   An element displayed using CSS grid layout showing an extra icon next to its name in the Inspect overlay
+:::image-end:::
+
+:::image type="complex" source="images/inspect-tool-flexbox-element.msft.png" alt-text="An element displayed using CSS flexbox showing an extra icon next to its name in the Inspect overlay" lightbox="images/inspect-tool-flexbox-element.msft.png":::
+   An element displayed using CSS flexbox showing an extra icon next to its name in the Inspect overlay
+:::image-end:::
+
```
captainbrosset, line 52: I would suggest having only 1 screenshot here. Either for grid or for flex, it doesn't really matter, but I think only 1 is sufficient to get the point across. We only want people to be aware that the icon you highlighted in red carries meaning.

microsoft-edge/devtools-guide-chromium/css/inspect.md
```diff
@@ -0,0 +1,82 @@
...
+The **Inspect** overlay shows the following information, depending on the type of element and the styles applied to it:
+
+* The name of the element
+* Its dimensions in pixels
+* Its color as a hexidecimal value and a color swatch
+* Its font settings
+* Its margin and padding in pixels
+
+If the element has been positioned using CSS grid or CSS flexbox you get different icons displayed next to its name.
```
captainbrosset, line 43: Minor rephrasing proposal: `If the element is positioned using CSS grid or CSS flexbox, a different icon is displayed next to the element's name`.

microsoft-edge/devtools-guide-chromium/css/inspect.md
```diff
@@ -0,0 +1,82 @@
...
+
+For example, in the demo page, the `Bad Contrast` button shows a warning icon next to the contrast value of 1.77. It also shows that the button is not focusable via keyboard as it is a `DIV` element with a class of button instead of a `BUTTON` element.
+
+:::image type="complex" source="images/inspect-tool-bad-contrast.msft.png" alt-text="Elements without sufficient contrast show a warning icon" lightbox="images/inspect-tool-bad-contrast.msft.png":::
+   Elements without sufficient contrast show a warning icon
+:::image-end:::
+
+## Inspecting non-accessible elements
+
+Elements that are covered by other elements or have the CSS property of `pointer-events: hidden` are not available to the **Inspect** tool. In the demo page, try to hover over the `Overlay Button` and you will see that the parent element - in this case `div.wrapper` is shown instead.
```
captainbrosset, line 63: The `Shift` trick doesn't work for elements that are covered by other elements. You can try it on this test page for example:

`data:text/html,<style>div{position:absolute;top:100px;left:100px;width:200px;aspect-ratio:1;background:red;}div+div{background:green;width:300px;}</style><div>1</div><div>2`

You won't be able to select div 1 even when holding shift.

The sentence further down (`If you want to inspect elements like that`) makes you think that holding `Shift` will work for both covered elements and `pointer-events:none` elements, but it won't.

So perhaps rephrase to `Elements that have the CSS style `pointer-events: none` are not available to the **Inspect** tool`.

(note how I also changed `hidden` to `none`, which is the correct CSS value).

microsoft-edge/devtools-guide-chromium/css/inspect.md
```diff
@@ -0,0 +1,82 @@
...
+
+If you want to inspect elements like that, you need to press `Shift` whilst hovering over elements. There is also a color overlay showing you that you are in an advanced selection mode.
+
+:::image type="complex" source="images/inspect-tool-with-shift.msft.png" alt-text="Pressing the Shift key whilst selecting elements on the page allows you to access elements more elements, even those with a pointer-events: none style setting" lightbox="images/inspect-tool-with-shift.msft.png":::
+   Pressing the Shift key whilst selecting elements on the page allows you to access elements more elements, even those with a pointer-events: none style setting
+:::image-end:::
+
+## Selecting the element and terminating Inspect mode
+
+If you click on any of the elements, the **Styles** tool shows the CSS applied to the element and the **Inspect** tool deactivates.
```
captainbrosset, line 77: Maybe also mention that the element becomes selected in the DOM tree?

captainbrosset (changes requested):

microsoft-edge/devtools-guide-chromium/css/inspect.md
codepo8: fixed

microsoft-edge/devtools-guide-chromium/css/inspect.md
codepo8: fixed

microsoft-edge/devtools-guide-chromium/css/inspect.md
codepo8: fixed

microsoft-edge/devtools-guide-chromium/css/inspect.md
codepo8: fixed

codepo8:
Looks great, changed one typo in the first paragraph.

codepo8 (commented):
Good re-write, makes it clearer

mikehoffms (approved):

mikehoffms:
Merging now per PR owner.

PR #1646, opened 2021-12-18
Cleaned up YAML fields

mikehoffms:
* Made tab title match top-of-page title.
* Put title: first.
* Put ms.date: last.
* Set ms.topic: to "conceptual" on all .md files except Include files.
* Removed "for Microsoft Edge Developer Tools" from title of DevTools articles.


vchapel (approved):
LGTM

PR #1676, opened 2022-01-13
Update Site Impacting changes

ericlaw1979:
Note 3-digit UA string.
Postpone cross-origin dialog change


ericlaw1979:
+ @mikehoffms

mikehoffms (approved):
@ericlaw1979 - Please techreview my commit; my rewording of the three-digit entry.  
What if the version # is 99, will `Edg/099` be returned instead of '`Edg/99`?  That might be an interesting example to provide.  
Will `Edg/099` ever be returned?  
Will `--enable-features=ForceMajorVersion100InUserAgent` always return `Edg/100`, or could it return `Edg/099`?

ericlaw1979 (commented):
Your edit looks fine.

ericlaw1979:
Edit looks fine. No, there's no case of a "Edg/099" version number.

The change here is that the version number is going from 99 to 100. The reason this is deemed a potentially-breaking change is only because of the fact that 100 has three digits while 99 has two. 

The command line flag does exactly what it says, it changes the major version to "100" even if it is really, say, 98.

PR #1692, opened 2022-01-24
node-fetch 3 was migrated to ESM, so updating import

captainbrosset:
The dependency bot updated our version of node-fetch to version 3+. This version was converted to ESM only package, so we can't use `require` anymore to import it.
We can, however, dynamically import it with the `import()` function.

Doing this will fix the `scripts/find-references-to-old-versions.js` script which would otherwise fail.


captainbrosset:
I merged this without review because this is a quick fix that was required to avoid breaking our script.

PR #1795, opened 2022-03-12
Popup refocus - F6 now supported, moves focus to next pane

mikehoffms:
**Rendered article section for review:**
https://review.docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/browser-features?branch=pr-en-us-1795#additional-keyboard-shortcuts-information
Find "F6", about the 18th row of the section's table.  Added a note in the F6 & Shift+F6 rows.

**Other changes:**
* Added h3 headings for other tables in that h2 section; those tables needed a skimmable heading/label.
*  Changed the table column 2 heading from "Windows" to "Shortcut", in most tables.

[Before](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/browser-features#additional-keyboard-shortcuts-information)

### DIFF
--- a/microsoft-edge/webview2/concepts/browser-features.md
+++ b/microsoft-edge/webview2/concepts/browser-features.md
@@ -114,7 +114,7 @@ To avoid such changes to your keyboard shortcuts, you can set `AreBrowserAcceler
 
 The following shortcuts are always turned off in WebView2, or are effectively turned off.  An asterisk (`*`) indicates that the shortcut isn't turned off, but the feature that it accesses is turned off, or the feature doesn't apply to WebView2.
 
-| Action | Windows |
+| Action | Shortcut |
 |:--- |:--- |
 | Add to Favorites | `Ctrl`+`D` |
 | Add All Tabs to Favorites | `Ctrl`+`Shift`+`D` |
@@ -131,8 +131,8 @@ The following shortcuts are always turned off in WebView2, or are effectively tu
 | Select Tab (1 - 8) | `Ctrl`+`(1-8)` |
 | Show Favorites Bar `*` | `Ctrl`+`Shift`+`B` |
 | Help | `F1` |
-| Focus Next Pane `*` | `F6`.  See [Support for F6 and Shift+F6](#support-for-f6-and-shiftf6). |
-| Focus Previous Pane `*` | `Shift`+`F6`.  See [Support for F6 and Shift+F6](#support-for-f6-and-shiftf6). |
+| Focus Next Pane `*` | `F6`.  Supported in windowed hosting mode, but not in visual hosting mode. |
+| Focus Previous Pane `*` | `Shift`+`F6`.  Supported in windowed hosting mode, but not in visual hosting mode. |
 | Reading View `*` | `F9` |
 | Focus Menu Bar | `F10` |
 | Show Identity Menu `*` | `Ctrl`+`Shift`+`M` |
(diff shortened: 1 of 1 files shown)


microsoft-edge/webview2/concepts/browser-features.md
```diff
@@ -129,8 +128,8 @@ The following shortcuts are always turned off in WebView2.  An asterisk (`*`) in
 | Select Tab (1 - 8) | `Ctrl`+`(1-8)` |
 | Show Favorites Bar `*` | `Ctrl`+`Shift`+`B` |
 | Help | `F1` |
-| Focus Next Pane `*` | `F6` |
-| Focus Previous Pane `*` | `Shift`+`F6` |
+| Focus Next Pane | `F6` |
+| Focus Previous Pane | `Shift`+`F6` |
```
mikehoffms, line 132: @tofuandeve @jasonstephen15 - Is Shift+F6 supported now?  Not mentioned in my work item & linked work item.  I would assume so; F6 & Shift+F6 are always "linked" as complementary in any UI.

microsoft-edge/webview2/concepts/browser-features.md
```diff
@@ -129,8 +128,8 @@ The following shortcuts are always turned off in WebView2.  An asterisk (`*`) in
 | Select Tab (1 - 8) | `Ctrl`+`(1-8)` |
 | Show Favorites Bar `*` | `Ctrl`+`Shift`+`B` |
 | Help | `F1` |
-| Focus Next Pane `*` | `F6` |
-| Focus Previous Pane `*` | `Shift`+`F6` |
+| Focus Next Pane | `F6` |
```
mikehoffms, line 131: @tofuandeve @jasonstephen15 - Should I add a note that F6 & Shift+F6 is currently only enabled in windowed hosting mode, not yet in visual hosting mode?

I'll look at .md file in VS Code to get a better view of where such note could go.

mikehoffms, line 131: @tofuandeve @jasonstephen15 - in my first commit, I merely removed the * from the F6 row and Shift+F6 row.
Should I remove the two rows, instead?  

See the statement above this table, in the published page
https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/browser-features#additional-keyboard-shortcuts-information - 
"The following shortcuts are always turned off in WebView2. An asterisk (*) indicates that the shortcut isn't turned off, but the feature that it accesses is turned off, or the feature doesn't apply to WebView2."

tofuandeve: @mikehoffms Yes, this is an important information, we should definitely add a note to explicitly call this out.

microsoft-edge/webview2/concepts/browser-features.md
tofuandeve: Yes both F6 and Shift+F6 are partially supported now in webview2 using windowed hosting mode

microsoft-edge/webview2/concepts/browser-features.md
tofuandeve: From the statement above, if we only remove the * then that would mean F6 is always turned off in WebView2 instead of saying we're partially supporting it now. But completely removing them would also not be very accurate since we are not fully enabling it. I was thinking maybe still leave them here with the * but specify that they are not supported in visual hosting mode? Would that be okay?

mikehoffms: @tofuandeve - yes that's the same conclusion i came to after reading the page more!  done; kept rows, kept *'s, added an h3 section and linked to it from the two rows.  Added other h3s that the page needed, as well.
@jasonstephen15

microsoft-edge/webview2/concepts/browser-features.md
```diff
@@ -129,8 +136,8 @@ The following shortcuts are always turned off in WebView2.  An asterisk (`*`) in
 | Select Tab (1 - 8) | `Ctrl`+`(1-8)` |
 | Show Favorites Bar `*` | `Ctrl`+`Shift`+`B` |
 | Help | `F1` |
-| Focus Next Pane `*` | `F6` |
-| Focus Previous Pane `*` | `Shift`+`F6` |
+| Focus Next Pane `*` | `F6`.  Supported in windowed hosting mode, but not in visual hosting mode. |
```
tofuandeve, line 139: This looks good to me but I'm not sure if these hosting modes are known concepts to end devs. Maybe add some more information like: `Supported in windowed hosting mode, but not in visual hosting mode (UWP WinUI 2 and Win32 C++ app with Visual Composition)`.  These are what we put in our docs here https://docs.microsoft.com/en-us/microsoft-edge/webview2/samples/webview2_sample_uwp so maybe that'll make it clearer for devs to look for information.

tofuandeve (approved):
Looks good to me 👍🏻

PR #1826, opened 2022-03-23
Switch WCO to fixed position

captainbrosset:
Certain browser/device pairs have elastic overscroll which means the page scrolls according to the scroll momentum, and can overscroll past its boundaries.

This makes positioning a title bar using WCO challenging as you do not want this title bar to overscroll.
Using fixed positioning solves this issue, and we should make sure this is what our documentation suggests.

### DIFF
--- a/microsoft-edge/progressive-web-apps-chromium/how-to/window-controls-overlay.md
+++ b/microsoft-edge/progressive-web-apps-chromium/how-to/window-controls-overlay.md
@@ -96,7 +96,7 @@ You can use these environment variables to position and size your app's title ba
 }
 ```
 
-Using `position:fixed;` makes sure your title bar does not scroll with the rest of the content and instead stays aligned with the window controls overlay.
+Using `position: fixed;` makes sure your title bar does not scroll with the rest of the content and instead stays aligned with the window controls overlay.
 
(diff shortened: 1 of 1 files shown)


microsoft-edge/progressive-web-apps-chromium/how-to/window-controls-overlay.md
```diff
@@ -88,14 +88,16 @@ You can use these environment variables to position and size your app's title ba
...
-    position: absolute;
+    position: fixed;
     left: env(titlebar-area-x);
     top: env(titlebar-area-y);
     height: env(titlebar-area-height);
     width: env(titlebar-area-width);
 }
 ```
 
+Using `position:fixed;` makes sure your title bar does not scroll with the rest of the content and instead stays aligned with the window controls overlay.
```
vchapel, line 99:
```suggestion
Using `position: fixed;` makes sure your title bar does not scroll with the rest of the content and instead stays aligned with the window controls overlay.
```

vchapel: Also, would it help to say "... fixed;` rather than `position: absolute;` ..." to make it more clear, because the old setting will no longer be visible?

captainbrosset: I'm not sure, because fixed and absolute are not the only 2 choices. So it's not a question of using either one or the other. So it feels odd to me to compare only to absolute, and not static, relative, and sticky for example.

PR #1978, opened 2022-05-27
Correct spelling of Notifications API name

jasonabullard:
Notifications API is defined as Notification in window object


captainbrosset (approved):
Thanks so much for catching this typo and fixing it.

PR #2112, opened 2022-08-11
FocusModeBash Update to console live expressions

JoshuaWeber:
Changes for Console Live Expressions. 

While I updated the screenshot and did some text updates for inaccuracies. I think this top section for Live Expressions needs to be re-written. It just does the same thing 3 times with I think little added value. It is also very complex and brittle relying on multiple demo websites as part of the walkthrough flow. 

**Rendered page:**
https://review.docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/console/live-expressions?branch=pr-en-us-2112


microsoft-edge/devtools-guide-chromium/console/live-expressions.md
```diff
@@ -48,15 +48,15 @@ To add a live expression:
 
    <!-- revise the step & the capture after it: -->
 
-1. In that same window, go to the webpage [Overview of DevTools](../overview.md), and then click the **Docs** button in the upper left.
+1. In that same window, go to the webpage [Overview of DevTools](../overview.md), and then click the **Docs** button in the upper right.
```
mikehoffms, line 51: we might delete this line later, but I think the intended "Docs" button is upper left, I don't see a "Docs" button in right.

image

mikehoffms, line 51:
```suggestion
1. In that same window, go to the webpage [Overview of DevTools](../overview.md), and then click the **Docs** button in the upper left.
```

captainbrosset: I also don't have the button in the upper right.
But, I agree with you both, using several demo sites to show the same thing, and especially using DMC, is not a good idea. So I'll log a task for us to rewrite this page. For now we can merge this as the new images look great. Thanks Josh.

captainbrosset (approved):

PR #2118, opened 2022-08-12
Update images on console

xiaofeihan1:
In this PR, we only change the images to focus mode.
We also need to change some text on the following places:
1. It seems that we don't show the errors in focus mode. see the below picture, the link is [here](https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/console/console-debug-javascript)
image
2. The same problem as above is in another doc. See [here](https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/console/#error-reports-and-the-console)


captainbrosset:
Thanks @xiaofeihan1, I'll rebase this branch to get rid of the build errors, and will make the required documentation changes.

captainbrosset:
@xiaofeihan1 I fixed the issues locally, but since I couldn't push back to your fork, I just moved your commits to another branch on this repo. Here it is: #2119. So I'll close this PR now. That doesn't mean we're not accepting your changes :) It's just that your work is now in the other PR.

PR #2133, opened 2022-08-18
Update and simplify the PWA getting started guide

captainbrosset:
The PWA getting started guide relies on the express library and the PWABuilder service worker library.
Because this is a getting started guide, I don't think this is a great idea. These libs are easy to use, but they also come with a lot of potential complexity that might confuse users only looking to understand what a PWA is.

This PR removes these dependencies by adding simple code users can copy/paste, updates the screenshots, and generally updates the content to be more in line with our writing style.

Rendered page for review: https://review.docs.microsoft.com/en-us/microsoft-edge/progressive-web-apps-chromium/how-to/?branch=pr-en-us-2133


htalat:
Should we add a screenshot of how the app ends up looking like?

captainbrosset:
> Should we add a screenshot of how the app ends up looking like?

Great idea, I'll add a screenshot

mhochk (approved):

JasonAndrewWriter (approved):

PR #2184, opened 2022-09-17
Show Sensors is listed as a Drawer (not Panel)

pankajparashar:
It would be useful to use a screenshot of the command menu illustrating the "Show Sensors" option, just like below,

image


captainbrosset (approved):

captainbrosset:
Thank you so much @pankajparashar. Merging this now.
As for adding a new screenshot: I agree with the proposal, but we also have to keep in mind that adding too many screenshots in procedure content like this makes it harder to read. I don't think the very first screenshot is needed, we could get rid of it and instead add the one you suggested.
We're currently doing a project where we're re-creating a lot of screenshots, so once we get to this particular page, we can revisit the screenshots on this page and do the change then.

captainbrosset:
I just merged this to our live branch, so the change will be live any minute now.

PR #2190, opened 2022-09-20
Fixed keyboard shortcut to open settings

pankajparashar:
Pretty sure on MacOS, the keyboard shortcut to open Settings is `Shift+?`. Didn't check windows, but most likely it should be the same.

### DIFF
--- a/microsoft-edge/devtools-guide-chromium/shortcuts/index.md
+++ b/microsoft-edge/devtools-guide-chromium/shortcuts/index.md
@@ -48,7 +48,7 @@ The following keyboard shortcuts are available in most DevTools panels.
 
 | Action | Windows/Linux | macOS |
 |---|---|---|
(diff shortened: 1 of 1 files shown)


captainbrosset:
@pankajparashar good point. On my keyboard, the `?` sign can only be accessed by pressing `Shift` first. So, the shortcut is indeed `Shift`+`?`. I wonder if there are other keyboard layouts where `?` can be accessed directly with no modifier.

@mikehoffms what do you think?

pankajparashar:
This is what the tooltip on the Settings icon in Edge shows the keyboard shortcut,

image

captainbrosset (approved):

captainbrosset:
Confirmed with Michael offline that we indeed need to add the Shift modifier in the doc. The ? key can only be accessed with Shift, on every keyboard.

captainbrosset:
This change has been merged to the live branch and should be available any minute now.

PR #2210, opened 2022-09-27
Updated with newest focus mode screenshots.

dan-wesley:
Added Application icon. Revised step 2 in Open the Cookies pane


captainbrosset (approved):
Thanks Dan. Great work on these.

I have one comment that I hope you can take into consideration for the following images (if any), but I don't want to have you redo these images because they are still an improvement over the previous ones.

My comment is about taking full window screenshots instead of cropped out versions.
The new images (and the old images too) do not show the window borders and corners that make it easy to identify this as an app window on desktop. They are cropped out to only show the insides of devtools.
This is something we want to move away from for 2 reasons:

1. Without a border, it's harder to see where an image starts and ends when rendered on learn.m.c. If the image has some white areas, those areas will blend right in with the white background of the article (I know there's a way to add borders in markdown on LMC, but we've moved away from this too).
2. It's harder for readers to get great context from a cropped-out image than it is from a full window screenshot. Is the cropped area supposed to be inside another window? If so where? We found that showing the entire window (which means including the titlebar, close/maximize/minimize buttons, and borders) produces very clean and easy to understand results.

Related to point 2: we're trying to capture the entire browser window as much as possible. Instead of undocking devtools in its own window, we keep it as it is by default (so on the right-hand side of the browser) and capture the full Edge window. This way it's really obvious to readers that we're talking about the tools that are inside of Edge. Also, most of the times the webpage that's loaded in the tab that devtools is currently inspecting is important too. Sometimes it's not, and in this case undocking devtools can be fine. But most of the times, it's just better to show the whole browser.

How to take full window screenshots easily:

1. Make your desktop background white.
2. Hide icons on your desktop.
3. Make Edge as default-looking as possible (no logged-in icon, no custom extensions, etc.)
4. Resize Edge and DevTools within it so they show the right amount of information.
5. Hide all other windows and keep just Edge.
6. Use Snipping Tool in "Window mode".
7. Capture the Edge window.

PR #2211, opened 2022-09-27
user/v danwesley/update focus mode screenshots for local storage article

dan-wesley:


dan-wesley:
@captainbrosset Hey there. Saw your comments on the cookies PR so will re-do the snapshots for this PR. It's not a huge deal to redo.

dan-wesley:
@captainbrosset Hey Patrick, redid the screenshots for the article showing the browser. Let me know if this is what you're looking for in the docs.

captainbrosset (changes requested):
Thanks this looks good.

A couple comments for even better screenshots next time (but please don't redo these ones, they're already an improvement over the current screenshots, no need to redo them, this is just for next time):

1. Use Snipping Tool's "Window Mode", un-maximize Edge so you can see its borders and rounded corners, make sure you have a white background and no other windows or icons are visible, and capture the Edge window. This is what we found produces the best results so far. Your screenshots do not show the Edge window's rounded corners and borders. Here is an example of a great screenshot, to compare with: https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/css/images/css-overview-tool-font-element.png

2. Keep the webpage as simple as possible. If the docs you are working with do not need any specific webpage, then use our default demo page: https://microsoftedge.github.io/Demos/demo-to-do/. And also disable Device Mode, as it makes the webpage more busy than it needs to be.

But again, this is for next time.

Now, for this PR, there is something to fix though. You deleted this image: `microsoft-edge/devtools-guide-chromium/media/storage-application-manifest-empty.msft.png` and this is causing a build error because `microsoft-edge/devtools-guide-chromium/storage/indexeddb.md` needs it.
Please revert the change.

PR #2225, opened 2022-10-02
Update consume-source-maps-from-azure.md

Meir017:
Add note about enabling source-maps in the preferences tab as well which is required in order for the source-maps to be fetched by the edge browser

added steps 2 and 3
image


captainbrosset:
Thanks so much for the added details. I think this is a helpful addition even if these options are enabled by default in DevTools. It's a good thing to ask people to double check them.
I'll merge this and just make a quick image fix so your new image looks like the other ones on this page. But other than this, great content. Thanks.

captainbrosset (approved):

captainbrosset:
Merged into `user/pabrosse/source-maps-setting` so I can modify the image in that branch too.

PR #2240, opened 2022-10-09
Fixed the case to represent the placeholder

pankajparashar:
I'm not sure what's the convention for representing placeholder text in documentations, but if it should match the case exactly, then you can merge this PR.


captainbrosset (approved):
Thanks for catching this and opening a PR!
We do want casing to match what's in the UI. I'll merge this.

captainbrosset:
This should be live in a bit.

PR #2246, opened 2022-10-11
users/v-jandrew/new-focus-mode images of reference

JasonAndrewWriter:
I redid the pr because I have the images saved on my laptop and it was easier to fix than fixing the merge.


microsoft-edge/devtools-guide-chromium/javascript/reference.md
```diff
@@ -36,7 +36,7 @@ To view and edit a JavaScript file:
 
 1. Select the **Sources** tool.
 
-1. In the **Navigator** pane, select your file, to open it in the **Editor** pane.
+1. In the **Navigator** pane, select your file, to open it in the **Ed0itor** pane.
```
mikehoffms, line 39:
```suggestion
1. In the **Navigator** pane, select your file, to open it in the **Editor** pane.
```

captainbrosset (changes requested):
Sorry about the delay. I took a look today, and I think we unfortunately need to redo these screenshots. There are a few things that need to be changed:

* The most important is that the **Quick View** section (at the bottom) is visible in all screenshots. Unless an article is specifically about **Quick View**, then we should hide it (by pressing `ESC`) before taking screenshots. It unnecessarily makes screenshots busier and harder to understand. But, also, **Quick View** is changing very soon, and the more we show it, the more screenshots we'll have to redo later, after it has changed.
* It feels strange to see different demo pages in every screenshot. It would be better to just stick to one demo page and use it throughout the article.
* Most screenshots show a few distracting pixels beyond the window borders. To achieve cleaner results, it's necessary to hide desktop icons and all other windows, and then choose a white background. This way, the Edge window sits on solid white, and it makes the final images much cleaner.
* The last 3 screenshots of the article have not been updated in this PR and still show non-focus mode UI and dark theme.

I would advise leaving this PR for the moment, and then picking it up again once **Quick View** has been changed. So I'd actually advise closing it for now, and we can then work on it again in a few time.

PR #2391, opened 2023-01-08
Update index.md

JaiganeshKumaran:


captainbrosset:
Thanks for spotting and fixing this. The change should be live on the site any minute now.

PR #2447, opened 2023-02-16
WebView2 JS Doc: Fix malformat

peiche-jessica:
Use `|-` instead of `>-` to fix missing linebreaks.

After the fix:
image

AB#43386857


peiche-jessica:
Somehow I couldn't add @mikehoffms as a reviewer.

Reezaali (approved):

PR #2551, opened 2023-04-18
Update winui2.md to note unsupported custom CSS cursors

david-risney:
I added a note that CSS cursors with custom images aren't supported in UWP apps. 

I've also updated the section at the end. It was talking about inaccessible interfaces, but WinUI2 uses WinRT which uses primarily our runtime classes not really our interfaces, so updated to the class names not interfaces. Also removed CoreWebView2Environment since that is accessible via the `CoreWebView2.Environment` property.

### DIFF
--- a/microsoft-edge/webview2/get-started/winui2.md
+++ b/microsoft-edge/webview2/get-started/winui2.md
@@ -378,6 +378,12 @@ On WinUI 2, transparency is achieved by setting the color to `00FFFFFF`.
 
 
 <!-- ------------------------------ -->
+#### Custom cursors
+
+On WinUI 2, you cannot use [CSS cursors](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) with a URL of an image as the cursor. You can use CSS cursors to change the cursor to a predefined cursor such as `cursor: wait;` or `cursor: crosshair;`, just not a value where you specify your own image via URL such as `cursor: url(https://example.com/cursor.png), pointer;`. See [CSS - cursor loaded from URL doesn't work](https://github.com/MicrosoftEdge/WebView2Feedback/issues/1925).
+
+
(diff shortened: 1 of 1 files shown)


microsoft-edge/webview2/get-started/winui2.md
```diff
@@ -377,6 +377,12 @@ See also:
 On WinUI 2, transparency is achieved by setting the color to `00FFFFFF`.
 
 
+<!-- ------------------------------ -->
+#### Custom cursors
+
+On WinUI 2, you cannot use [CSS cursors](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) with a URL of an image as the cursor. You can use CSS cursors to change the cursor to a predefined cursor such as `cursor: wait;` or `cursor: crosshair;`, just not a value where you specify your own image via URL such as `cursor: url(https://example.com/cursor.png), pointer;`. See [CSS - cursor loaded from URL doesn't work](https://github.com/MicrosoftEdge/WebView2Feedback/issues/1925).
```
mikehoffms, line 383:
```suggestion
On WinUI 2, you cannot use [CSS cursors](https://developer.mozilla.org/docs/Web/CSS/cursor) with a URL of an image as the cursor. You can use CSS cursors to change the cursor to a predefined cursor such as `cursor: wait;` or `cursor: crosshair;`, just not a value where you specify your own image via URL such as `cursor: url(https://example.com/cursor.png), pointer;`. See [CSS - cursor loaded from URL doesn't work](https://github.com/MicrosoftEdge/WebView2Feedback/issues/1925).
```

mikehoffms (changes requested):
Writer/Editor pass

mikehoffms (approved):

nishitha-burman (approved):

PR #2553, opened 2023-04-18
Runtimeclass names for API limitations

david-risney:
Switch to using the WinRT runtime class names for the inaccessible APIs rather than the COM interface names which aren't really available through our WinRT APIs. This is a subset of the change just made to the WinUI2 docs.


david-risney:
(@mikehoffms fyi)

mikehoffms (approved):

nishitha-burman (approved):

PR #2587, opened 2023-05-11
Update how-to-detect-win11.md

erik-anderson:
Call out the source of the version number on Windows.

AB#44601627


majenny (approved):
LGTM!

PR #2605, opened 2023-05-25
Add supplement for attaching to uwp app in webdriver.md

yanlingwang23:
Add supplement that attaching webdriver to UWP app is only supported on Canary now.


microsoft-edge/webview2/how-to/webdriver.md
```diff
@@ -230,6 +230,10 @@ For more info about the `DebuggerAddress` property of the `EdgeOptions` object,
 ### For UWP apps, attach Microsoft Edge WebDriver to WebView2 with Remote Tools
 
 1. Complete the steps described in [Remote debugging UWP apps with Remote Tools for Microsoft Edge](./remote-debugging.md).
+
+   As for now, Canary is the only supported version to test, please set the enviroment variable      `WEBVIEW2_RELEASE_CHANNEL_PREFERENCE=1` to make sure the Webview2 version is Canary.
```
mikehoffms, line 234:
```suggestion
   The Canary preview channel of Microsoft Edge is currently the only supported version of WebView2.  Set the enviroment variable `WEBVIEW2_RELEASE_CHANNEL_PREFERENCE=1` to ensure that the WebView2 version is Canary.
```

mikehoffms: @yanlingwang23 - do we need to specify a version number, Canary 115?

yanlingwang23: The code change is checked-in in 115.0.1849.0. I've also checked the available version for webdriver listed in the https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/#installation and all the recent version listed under Canary included the change.

So as long as user is using the Canary preview channel (the version of it is above or equal to 115) and downloaded the webdriver for the specific version 115.0.1849.0 or above, they should be able to test it. 

Also the version of webdriver that user downloaded should be consistent with the browser version, for instance, both using 115. (This part is already covered in the https://learn.microsoft.com/en-us/microsoft-edge/webdriver-chromium/?tabs=c-sharp, which is also linked at the beginning of this file as 'Webdriver overview', so I feel like we might not need to address that again here).

And as time goes on, the change will finally get into stable channel and at that point we can remove this part.

mikehoffms: @yanlingwang23 - pls incorp my Suggested Change if it looks good.
@zoherghadyali - can we close this thread after the Suggested Change? pls see Yanling's explanation.

mikehoffms (approved):

captainbrosset (approved):

PR #2640, opened 2023-06-14
Move images to be next to the markdown files that refer to them

captainbrosset:
After maintaining this doc set for a while, we've found that it's much better for each markdown file to have its own images rather than share images across multiple articles.

Articles diverge over time, and even if 2 articles may need the same image at some point, there's a high chance they'll be updated independently and will need different images later.

This PR (which I created by running various scripts locally) moves all images to folders next to the md files that refer to them.
For example, if a md file is named `article.md`, then a new folder named `article-images` is created next to it and all images are moved into it.

There are a few exceptions to this:

* the devtools icons remain in a general `media` folder under `devtools-guides-chromium`
* the creative commons license logo remains in a top-level `media` folder too, and so does the experimental flag banner.


Reezaali (approved):

Reezaali:
Reviewed all MD files on staging. All good.

mikehoffms (approved):

PR #2679, opened 2023-07-01
Update index.md

FlyingBlueMonkey:
Grammer correction.  Removed unnecessary "when" from "during when installation"

AB#45325081


captainbrosset (approved):
Thanks!

PR #2873, opened 2023-10-17
JSON viewer: removing the experimental flag and adding a note about `window.open`

captainbrosset:
The JSON viewer has been available in Edge for a couple of releases without the need to set a flag. So this PR removes the corresponding part of the docs.
Also, the JSON viewer doesn't work in windows opened with `window.open`, so adding a note about this.

* Before: https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/json-viewer/json-viewer
* After: https://review.learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/json-viewer/json-viewer?branch=pr-en-us-2873

AB#47103759


microsoft-edge/devtools-guide-chromium/json-viewer/json-viewer.md
```diff
@@ -85,7 +69,7 @@ To view invalid JSON data:
 
 1. Type a URL that returns invalid JSON data in the address bar. For example, use [this sample invalid JSON response](https://microsoftedge.github.io/Demos/json-dummy-data/missing-colon.json). You can find other invalid samples of JSON responses at [MicrosoftEdge / Demos > json-dummy-data](https://microsoftedge.github.io/Demos/json-dummy-data/#invalid-json).
 
-1. Microsoft Edge detects that the file contains invalid JSON data, highlights it, but does not reformats it.
+1. Microsoft Edge detects that the file contains invalid JSON data, highlights it, but does not reformat it.
```
mikehoffms, line 72:
```suggestion
1. Microsoft Edge detects that the file contains invalid JSON data and highlights the invalid JSON data, but doesn't reformat it.
```
* Prefer contraction (except where emphasis is needed), per brand voice.
* The original comma-delimited list of phrases feels non-parallel, maybe due to uneven length of phrases.
* Suggest minimizing the use of "it", which is an indirection pointer that the user has to resolve.

mikehoffms (changes requested):
Writer/Editor review

microsoft-edge/devtools-guide-chromium/json-viewer/json-viewer.md
captainbrosset: The word "highlight" here is a reference to the intro paragraph where we say: "The JSON syntax is highlighted with different colors". So, it means that different parsed tokens of the JSON data are rendered in different colors. We could replace every instance of "highlight" with "syntax highlight" if that helped. What do you think?

captainbrosset: Pushed a commit to hopefully clarify this.

microsoft-edge/devtools-guide-chromium/json-viewer/json-viewer.md
mikehoffms, line 1: This article needs more synonym lookup/ index entries in toc.yml:  
that is,
        - name: View formatted JSON
          href: devtools-guide-chromium/json-viewer/json-viewer.md
          displayName: JSON viewer
should be:
          displayName: JSON viewer, view JSON

mikehoffms (approved):
Writer/Editor review

PR #2914, opened 2023-11-11
Fixed labels to match the actual options

pankajparashar:
AB#47570678


captainbrosset (approved):
Thanks for finding and fixing this!

PR #2961, opened 2023-12-05
Revise Distribution article per 120

mikehoffms:
Rendered article sections for review: 
https://review.learn.microsoft.com/microsoft-edge/webview2/concepts/distribution?branch=pr-en-us-2961#details-about-the-fixed-version-runtime-distribution-mode - step 6, "Starting with Fixed Version 120, " & removes "PlayReady" mentions

AB#48046869

### DIFF
--- a/microsoft-edge/webview2/concepts/distribution.md
+++ b/microsoft-edge/webview2/concepts/distribution.md
@@ -6,7 +6,7 @@ ms.author: msedgedevrel
 ms.topic: conceptual
 ms.prod: microsoft-edge
 ms.technology: webview
-ms.date: 08/21/2023
+ms.date: 12/05/2023
 ---
 # Distribute your app and the WebView2 Runtime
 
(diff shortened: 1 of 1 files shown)


microsoft-edge/webview2/concepts/distribution.md
```diff
@@ -348,13 +345,16 @@ To use the Fixed Version distribution mode:
       icacls {Fixed Version path} /grant *S-1-15-2-1:(OI)(CI)(RX)
       ```
 
-      PlayReady should be working now on the user's device.
-
-   1. To confirm that PlayReady is installed correctly, in the **Security** tab of the **Fixed Version** folder, make sure permissions are granted for `ALL APPLICATION PACKAGES` and `ALL RESTRICTED APPLICATION PACKAGES`, as shown below:
+   1. Confirm that permissions are granted for `ALL APPLICATION PACKAGES` and `ALL RESTRICTED APPLICATION PACKAGES` in the **Security** tab of the **Fixed Version** folder, as shown below:
```
mikehoffms, line 348: note:
https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/can-may -- "Don't use _may_, which might be interpreted as providing permission"
repo uses 'may' in some places, but usually not w/ the implication that we are granting the reader permission 

here, sidestep can vs may, by standard direct instruction verb

nishitha-burman (approved):

PR #2965, opened 2023-12-08
Replace "Large request rows" with "Big request rows"

pankajparashar:


captainbrosset (commented):
Thanks Pankaj. This looks great. 
We use Windows for all of the screenshots on the site. So let me add a commit with a couple of new screenshots real quick.

PR #2996, opened 2023-12-27
Added a tip to autocomplete styles using values

pankajparashar:
Rendered article section for review:
https://review.learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/css/reference?branch=pr-en-us-2996#adding-an-inline-css-declaration-to-an-element -- new paragraph like:

Alternatively, enter the value in the property field, and DevTools will then suggest a list of matching property: value pairs to select from. For example, if you enter bold in the property field, ...

AB#48295949

### DIFF
--- a/microsoft-edge/devtools-guide-chromium/css/reference.md
+++ b/microsoft-edge/devtools-guide-chromium/css/reference.md
@@ -216,8 +216,7 @@ To add an inline declaration:
 
 1. Enter a valid value for that property and press **Enter**.  In the **DOM Tree**, a `style` attribute has been added to the element.
 
-> [!TIP]
-> Alternatively, you can also enter the value in the property field, and DevTools will suggest a list of matching **property: value** pairs to select from. For example, if you enter `bold` in the property field then DevTools will suggest `font-weight: bold` and `font-weight: bolder` as the possible rules. Press **Enter** to apply the rule.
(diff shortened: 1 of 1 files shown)


microsoft-edge/devtools-guide-chromium/css/reference.md
```diff
@@ -216,6 +216,9 @@ To add an inline declaration:
 
 1. Enter a valid value for that property and press **Enter**.  In the **DOM Tree**, a `style` attribute has been added to the element.
 
+> [!TIP]
+> Alternatively, you can also enter the value in the property field, and DevTools will suggest a list of matching **property: value** pairs to select from. For example, if you enter `bold` in the property field then DevTools will suggest `font-weight: bold` and `font-weight: bolder` as the possible rules. Press **Enter** to apply the rule.
```
mikehoffms, line 220:
```suggestion
Alternatively, enter the value in the property field, and DevTools will then suggest a list of matching **property: value** pairs to select from. For example, if you enter `bold` in the property field, DevTools suggests `font-weight: bold` and `font-weight: bolder` as the possible rules. Press **Enter** to apply the rule.
```
* prefer present not future tense, per style guide
* Tip is deprecated per https://review.learn.microsoft.com/en-us/help/platform/markdown-reference?branch=main#alerts-note-tip-important-caution-warning
* I'll test this UI, read this article section, and possible enter a later rewrite comment.
* Omit "then" in an "If" sentence, per https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/i/if-vs-whether-vs-when

PR #3090, opened 2024-03-07
Change the search scope to narrow down to edge developer

captainbrosset:
Rendered docs for review: 
* https://review.learn.microsoft.com/microsoft-edge/developer/?branch=pr-en-us-3090

Changing the search scope to **Microsoft Edge Developer** would better match our docset, and return fewer, but more useful, results when using the search tool.

This would make the **Filter by title** box show a more relevant last option:

image

And would hopefully narrow down the results to our docset only. Currently, the results are for Microsoft Edge, and include a lot 
of non-developer documentation.

My only uncertainty is whether **Microsoft Edge Developer** is actually a scope that can be used on Learn. Currently, it returns nothing:

image

AB#49334958

### DIFF
--- a/microsoft-edge/docfx.json
+++ b/microsoft-edge/docfx.json
@@ -35,7 +35,7 @@
       "breadcrumb_path": "/microsoft-edge/breadcrumbs/toc.json",
       "feedback_system": "GitHub",
       "feedback_github_repo": "MicrosoftDocs/edge-developer",
-      "searchScope": ["Microsoft Edge Developer"],
+      "searchScope": ["Microsoft Edge Developer", "Microsoft Edge"],
       "titleSuffix": "Microsoft Edge Developer documentation",
       "uhfHeaderId": "MSDocsHeader-MSEdge"
     },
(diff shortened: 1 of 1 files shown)


mikehoffms (commented):
"formatted" returns 0 hits, although repo contains 48 hits in 23 files (mostly articles):

https://review.learn.microsoft.com/microsoft-edge/developer/?branch=pr-en-us-3090

image

captainbrosset:
> My only uncertainty is whether Microsoft Edge Developer is actually a scope that can be used on Learn. Currently, it returns nothing.

I now understand how this works. You can use new search scopes in docfx.json. The first time a new search scope is introduced it returns nothing, and a few days later, it starts working.
So, the idea is to introduce the new search scope while also keeping the old one, so that results are still returned. And after a few days, we can remove the old one if we want.

The order matters too: `"searchScope": ["Microsoft Edge Developer", "Microsoft Edge"]` means that **Microsoft Edge** results are returned first, which is good because the newly introduced **Microsoft Edge Developer** scope won't work just yet.
Once it does work, we could either remove **Microsoft Edge** or reverse the order: `"searchScope": ["Microsoft Edge", "Microsoft Edge Developer"]`.
This way, developer docs results would be returned first, but we'd still have general edge results too, in case no matches are found.

captainbrosset:
Testing might be hard, I don't think the site search tool works on the preview server.

mikehoffms (approved):
Approved PR

PR #3097, opened 2024-03-11
Reverse the order in `searchScope` to prioritize developer docs in search results

captainbrosset:
Now that the new **Microsoft Edge Developer** search scope has been introduced and returns results (see [this search query](https://learn.microsoft.com/en-us/search/?scope=Microsoft%20Edge%20Developer&terms=capture%20screenshot)), it's time to reverse the order of the `searchScope` array in docfx.json. 

Making **Microsoft Edge Developer** be last in the array means that results that belong to our docset will be returned first.


mikehoffms (approved):
Approved

PR #3174, opened 2024-05-22
Remove unneeded request parameter from LAF support shortlink

captainbrosset:
Turns out this second parameter is not needed. I'll file an internal issue to get the template used by the reference docs in this PR updated too.

AB#51192189


mikehoffms (approved):
Approved per Writer/Editor review.

PR #3317, opened 2024-11-25
Add note describing discontinuing Hololens support

champnic:
Hololens 2 is slowly being discontinued/deprecated, which means no more updates or technical support for the platform. Add a note to our support page indicating this, and that there is a risk that the applications may break in the future.


mikehoffms (approved):
LGTM - Writer/Editor

PR #3382, opened 2025-02-15
WebView2: VS Code debug config corrections, and source map loading description

mikosg:
**VS Code debugging:**
- `"useWebView": true` property is not allowed for `"request": "attach"` type configs. For attaching configs, `useWebView` is only needed for UWP apps, and accepts an object instead of boolean, where the debug pipe shall be specified.
- `--remote-debugging-port=9222` command line argument is also needed for  `"request": "launch"` type configs when custom `runtimeExecutable` is used. It can be supplied either via environment variable or via registry.

**Source map loading:**

WebView2 does not load source maps referenced by resources which were loaded using either `WebResourceRequested` event or virtual host name mapping.  This is a [known limitation](https://github.com/MicrosoftEdge/WebView2Feedback/issues/961?fbclid=IwZXh0bgNhZW0CMTEAAR2ipcsx9oeFkv51KZIKVgLvnIyk2x1SkoxQOdfl6UamLQOgfeMJVIME4SY_aem_jwSPdIAvI7ihPWuNph3SXA) which I also verified, and worth to be documented here along with its possible workarounds.

### DIFF
--- a/microsoft-edge/webview2/concepts/working-with-local-content.md
+++ b/microsoft-edge/webview2/concepts/working-with-local-content.md
@@ -353,7 +353,7 @@ When loading local content via a virtual host name mapping, you are mapping a vi
 <!-- ---------- -->
 ###### Additional web resources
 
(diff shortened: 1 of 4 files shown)


mikosg:
@microsoft-github-policy-service agree

mikosg:
Found some additional information which shall be corrected in the PR

PR #3678, opened 2025-12-31
Fix relnotes script failure when there's no Edge-only OTs

captainbrosset:
Need to merge this PR relatively quickly if we want to publish the next beta web platform release notes in time.

Files for review:
* `scripts/web-platform-release-notes.js`
   * Previously, the script assumed that there's always at least one Edge-only origin trial to fetch from https://developer.microsoft.com/en-us/microsoft-edge/origin-trials/trials.
When that's not the case, the script times out and the build fails.
   * Fixed this, by returning an empty array if the wait for Edge-only OTs takes more than 15 seconds.

AB#60644884

### DIFF
--- a/scripts/package-lock.json
+++ b/scripts/package-lock.json
@@ -12,8 +12,7 @@
         "@actions/github": "^6.0.0",
         "@octokit/auth-action": "^5.1.1",
         "@octokit/request": "^9.2.1",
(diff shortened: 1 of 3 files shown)


mikehoffms (approved):
LGTM - Writer/Editor

PR #3880, opened 2026-07-25
Document how the virtual host name affects navigation time

Burakuslendera:
In PR's offline group chat:
This PR is probably needed, and code base changes might also be needed - investigating.

Rendered article sections for review:

* **Using local content in WebView2 apps** > **Considerations for loading local content by using virtual host name mapping** > **Choosing a virtual host name**
   * `/webview2/concepts/working-with-local-content.md`
   * Internal preview: https://review.learn.microsoft.com/microsoft-edge/webview2/concepts/working-with-local-content?branch=pr-en-us-3880#choosing-a-virtual-host-name
   * External preview: https://github.com/Burakuslendera/edge-developer/blob/virtual-host-name-resolution-delay/microsoft-edge/webview2/concepts/working-with-local-content.md#choosing-a-virtual-host-name
   * Planned live: https://learn.microsoft.com/microsoft-edge/webview2/concepts/working-with-local-content#choosing-a-virtual-host-name
   * New section.

* **Using local content in WebView2 apps** > **Considerations for loading local content by handling the WebResourceRequested event** > **Choosing a host name for the origin**
   * `/webview2/concepts/working-with-local-content.md`
   * Internal preview: https://review.learn.microsoft.com/microsoft-edge/webview2/concepts/working-with-local-content?branch=pr-en-us-3880#choosing-a-host-name-for-the-origin
   * External preview: https://github.com/Burakuslendera/edge-developer/blob/virtual-host-name-resolution-delay/microsoft-edge/webview2/concepts/working-with-local-content.md#choosing-a-host-name-for-the-origin
   * Planned live: https://learn.microsoft.com/microsoft-edge/webview2/concepts/working-with-local-content#choosing-a-host-name-for-the-origin
   * New section.

AB#63298553

---
## What this PR adds

This PR adds two short subsections on choosing the host name for a synthetic origin, one under each approach on this page that uses one: 
* The virtual host name mapping approach.
* The `WebResourceRequested` approach.

## Why

The host name is resolved by the network stack before the mapping is applied, so a name that doesn't resolve costs a DNS timeout on every navigation.
The page currently says nothing about this, and its own example maps `demo`, which is such a name.

The times are measured on WebView2 runtime 150.0.4078.83, on an app that serves its content through `WebResourceRequested` rather than folder mapping, so nothing needs the name resolved at all. 
A NetLog capture names the span:

```
BEGIN  HOST_RESOLVER_MANAGER_JOB  host="app.local:443"  types=[A,AAAA]
END    HOST_RESOLVER_MANAGER_JOB
                                   2.007 s
```

That job covers exactly the interval between the main document being handed to `put_Response` and the renderer requesting the first subresource. 
Changing only the host name collapses it:

| host name | document served to first subresource | launch to window visible |
| --- | --- | --- |
| `app.local` | 2.012 to 2.041 s, seven runs | 2419 to 2543 ms |
| `app.localhost` | 11 to 79 ms | 495 to 508 ms |

The distinction the new text draws is the one that matters in practice. 
Reserved TLDs such as `.example`, `.test` and `.invalid` don't help, because RFC 2606 reserves them only against registration; the resolver still queries the network and still waits for the answer. 
RFC 6761 pins `.localhost` to the loopback interface, so the resolver answers it without a query. 
Renaming to `.test` was measured first, and changed nothing.

## Related Issue

[2 seconds delay is seen with Page navigation using WebView2 & SetVirtualHostNameToFolderMapping](https://github.com/MicrosoftEdge/WebView2Feedback/issues/2381) - Issue 2381 in the "MicrosoftEdge / WebView2Feedback" repo
This Issue reports the same delay for virtual host name mapping.
The Issue has been open since 2022. 
The measurement above extends it to the `WebResourceRequested` path, which suggests that the cost belongs to the synthetic origin itself, rather than to either API.


Burakuslendera:
@microsoft-github-policy-service agree

mikehoffms:
In PR's offline group chat: 
This PR is probably needed, and code base changes might also be needed - investigating.

krbharadwaj (commented):
I believe that the mapped resource request itself does not go through normal network/DNS resolution. Once  EmbeddedBrowserProxyURLLoaderFactory  finds the host mapping, it bypasses  target_factory_ , converts the mapped path to  file:// , and uses the file URL loader.

Is it some navigation path, which is causing the DNS resolution for virtual host requests as well?

Burakuslendera:
@krbharadwaj. You are right to distinguish the mapped resource load from other navigation work. My earlier measurement used `WebResourceRequested`: the `.local` delay and concurrent resolver job were real observations, but that trace did not identify which request initiated name resolution. I overgeneralized those observations into a DNS-before-mapping explanation for virtual host name mapping. The current PR wording makes a stronger causal claim than the evidence supports.

I ran a separate minimal test on WebView2 Runtime 153.0.4234.48, using fresh user data folders and NetLogs for each run. With a `.local` host, top level navigations using `SetVirtualHostNameToFolderMapping` and `WebResourceRequested` completed with `IsSuccess=true` in 1.323 s and 1.342 s, respectively. A virtual host mapped `.localhost` control completed in 34 ms. In both `.local` traces, the target host resolver job (~1.3 s, `ERR_NAME_NOT_RESOLVED`) was linked through `source_dependency` to an `HTTP_STREAM_JOB_CONTROLLER` for the navigation URL marked `is_preconnect: true`.

This points to navigation associated preconnect as a possible source of the resolver activity in this fixture. It does not prove that the mapped resource fetch waits for DNS, that preconnect alone caused the navigation delay, or that it explains the older report in [MicrosoftEdge/WebView2Feedback#2381](https://github.com/MicrosoftEdge/WebView2Feedback/issues/2381). 

Is that preconnect path expected for a mapped origin, and could the WebView2 team investigate whether it affects navigation timing? Please advise whether this PR should document the observed `.local` caveat while the behaviour is investigated.

krbharadwaj:
> Kriti Bharadwaj (@krbharadwaj). You are right to distinguish the mapped resource load from other navigation work. My earlier measurement used `WebResourceRequested`: the `.local` delay and concurrent resolver job were real observations, but that trace did not identify which request initiated name resolution. I overgeneralized those observations into a DNS-before-mapping explanation for virtual host name mapping. The current PR wording makes a stronger causal claim than the evidence supports.
> 
> I ran a separate minimal test on WebView2 Runtime 153.0.4234.48, using fresh user data folders and NetLogs for each run. With a `.local` host, top level navigations using `SetVirtualHostNameToFolderMapping` and `WebResourceRequested` completed with `IsSuccess=true` in 1.323 s and 1.342 s, respectively. A virtual host mapped `.localhost` control completed in 34 ms. In both `.local` traces, the target host resolver job (~1.3 s, `ERR_NAME_NOT_RESOLVED`) was linked through `source_dependency` to an `HTTP_STREAM_JOB_CONTROLLER` for the navigation URL marked `is_preconnect: true`.
> 
> This points to navigation associated preconnect as a possible source of the resolver activity in this fixture. It does not prove that the mapped resource fetch waits for DNS, that preconnect alone caused the navigation delay, or that it explains the older report in [MicrosoftEdge/WebView2Feedback#2381](https://github.com/MicrosoftEdge/WebView2Feedback/issues/2381).
> 
> Is that preconnect path expected for a mapped origin, and could the WebView2 team investigate whether it affects navigation timing? Please advise whether this PR should document the observed `.local` caveat while the behaviour is investigated.

Yes, we should investigate on the navigation preconnect delay with virtual host mapping scenarios. Ideally I feel that if pre-connect is not needed for VH scenario, then we can skip or fix that part in code itself, instead of suggesting any config changes at WV2 app side. Can we please put this document update on hold and create a work item for webview2 team to fix this issue?

